PR #39, opened 2026-06-06
feat(run): drop Docker, run the shell on the host (allow-all by default, SC_TRUST off-switch)

jedbjorn:
## Why

The container existed for one reason: to make **allow-all permissions** safe by being a boundary. But the agent and the services it operates — pm2, the host app — ended up on **opposite sides of that wall**. The shell couldn't restart, couldn't `curl` the running app, couldn't read logs, so problems became a fog. Every workaround (mount the pm2 socket, `--network host`) just punches holes in the fence *while keeping the Docker tax*. Co-locating the agent with what it operates is the whole job — so the box goes.

**The trade, stated plainly:** allow-all is a *prompting* setting, not a *filesystem fence*. Docker was the fence; without it, allow-all on the host means bash can reach anything the user can. We're trusting the model + the repo, not a wall. Edits stay soft-scoped to the repo; bash is open. This is a deliberate, accepted trade for a single-operator dev substrate.

## What changed

- **`sc`** — `enter`/`boot` exec the harness directly on the host (no `docker exec`). `launch`/`down`/`logs` manage a **backgrounded review-GUI** process via a pidfile under `.super-coder/run/` instead of a container. Dropped the docker plumbing (`dcheck`/`dcreds`/`duser`/`dbuild`), the `build` command, and the image.
- **Deleted** `Dockerfile` + `.dockerignore`; pruned from `update.py`'s engine manifest and the make aliases.
- **Decoupled allow-all from the (now absent) container:** `SC_SANDBOX` → **`SC_TRUST`**, default **ON** (host is the trust boundary). Off-switch: `SC_TRUST` in `{0,false,no,off}` keeps normal prompts — so a cautious fork can dial it back. `apply_sandbox` → `apply_trust`; adapter `sandbox` key → `trust`.
- **claude trust target** moves to `.claude/settings.local.json` with `defaultMode=bypassPermissions` — real allow-all (`acceptEdits` did **not** cover bash, confirmed against the docs), gitignored by convention, and it never clobbers a fork's tracked `.claude/settings.json`. opencode unchanged (`bash:allow` into the already-gitignored emitted `opencode.json`).
- **`doctor`** now checks host readiness (harness installed + login) instead of running a docker preflight. README / help / install messages updated to the host model.

## Verification

- `sh -n sc` + `py_compile` on all touched scripts + adapter JSON validate — pass.
- Headless `./sc verify`: boots, writes the **gitignored** `.claude/settings.local.json` (`bypassPermissions`), prints `→ trust: allow-all permissions`.
- `SC_TRUST=0`: no file written, prints `→ trust: OFF — normal permission prompts`.
- GUI lifecycle: `launch` (background) → `health` responds → idempotent `launch` ("already up") → `down` → `down` again ("not running"). All pass.

## Propagation

Engine files (`sc`, `.super-coder/*`) reach forks via `./sc update`. Existing forks may retain a now-orphaned `Dockerfile` (harmless — nothing references it) until manually cleaned.


jedbjorn:
Discarding — keeping the Docker isolation. The pm2/app-reachability friction is the accepted cost; manual paste back-and-forth is preferable to giving up the boundary. Work preserved at commit c691a033fe05d66b86c1a68f2cc8df41c5d0087b on the (now-deleted) branch if ever resurrected.

PR #131, opened 2026-06-17
docs(spec): host-side VM broker for sandboxed forks

jedbjorn:
## What

Specs a **host-side VM broker** so the shells that hold `windows_devkit` can actually drive the Windows Test VM from inside the super-coder sandbox container.

## Why

Measured from the live `sc-dos-arch` container:
- container → VM:22 **times out** (no route across libvirt NAT)
- container has **no ssh/scp/virsh** (python3 only)
- container → host gateway is **refused, not dropped** (route exists; host services bind 127.0.0.1)
- the fork repo is **bind-mounted** both sides → a unix socket can live there; `~/.ssh` is **not** mounted

So 'just pass the SSH creds into the container' fails on three walls — and even past them, `reset`/`capture` are host `virsh` ops a container can't (and shouldn't) run.

## Design

A host process holds the key + libvirt access and exposes `push/exec/capture/reset` (+ relocated `validate`) over a **unix socket in the bind-mounted repo**. `windows_devkit` curls the socket; the key never enters the fork; `virsh` runs host-side. Mirrors dos-arch's credential-broker precedent and is the generic test-target seam the [Windows Test VM design](docs/windows-test-vm.md) deferred.

Includes the `reset --running` note (offline snapshot, `invtsc` blocks live snapshots), phasing (MVP exec+reset → full → generalize), and open questions (where the #129 validate endpoints actually run, socket vs TCP, broker auth).

Doc only — no code.

### DIFF
--- a/docs/windows-vm-broker.md
+++ b/docs/windows-vm-broker.md
@@ -1,0 +1,203 @@
+---
+title: Windows VM Broker — Spec
+tags: [super-coder, design, testing, windows, broker]
+date: 2026-06-17
+project: super-coder
+purpose: Host-side broker so sandboxed forks can drive the test VM
+---
+
+# Windows VM Broker — Spec
+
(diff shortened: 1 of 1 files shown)


jedbjorn:
Superseded by #132, which carries the spec commit plus the full broker implementation.

PR #179, opened 2026-06-26
fix(gui): harden publish + add rolling event log & gh-auth docs

jedbjorn:
## Problem

The GUI **snapshot** and **publish** buttons stopped working. Root cause (reproduced live in the dos-arch fork): publish assumed it always starts on `main` with a clean tree. A prior run had left the working tree stranded on the ephemeral `sc_gui_content` branch, and from there publish could never recover:

1. `git_publish()` serialized first → dirtied `.sc-state/content.sql` on the *wrong* branch.
2. `_publish_content()` ran `git branch -D sc_gui_content` **while on it** → fails (`used by worktree`); returncode ignored, falsely logged `(dropped stale …)`. Then `checkout -b` → fails (already exists).
3. `_land_on_base()` ran `checkout main` → fails (dirty regenerated content would be overwritten) → stays stranded.
4. Next publish repeats. **Self-perpetuating.** Snapshot worked in isolation but every run re-dirtied the wrong branch.

## Fix

Exploits that **everything publish touches is regenerated from the live DB** — so working-tree edits to those paths are disposable.

- **`_prepare_branch()`** (new) runs *before* the snapshot: restores regenerable dirt, lands on a clean base, drops a stale local branch (now safe + returncode-checked), recreates from base. **Refuses** rather than clobbers when *unrelated* files are dirty; warns on orphaned commits.
- **`git_publish()`** reordered to prepare → snapshot/render → commit, wrapped so any raise still lands back on base.
- **`_land_on_base()`** restores regenerable dirt before `checkout` so cleanup can't re-strand.
- **`run_snapshot_render()`** raises on snapshot/render failure instead of committing stale flat files.
- **`_restore_regenerable()`** resets only concrete dirty paths — `git checkout -- <pathspec>` is fatal on any non-matching pathspec (forks lack some), which would abort the whole reset. (Caught by a test before shipping.)

## Tests

`tests/test_publish_prepare.py` — stranded-tree recovery, clean creation, refuse-on-unrelated-dirty, and both helpers (real git, tmp repo, stdlib unittest). Full suite: **107 passed**.


jedbjorn:
Extended this PR with **observability + the gh-auth gap** that the live publish incident exposed (we couldn't tell what the publish did, and host-run servers couldn't see a web-authed `gh`):

**Rolling webapp event log** — one file, last **20 end-to-end events**, JSON-per-line.
- `log_event()` / `read_log()` in `server.py`; wired into `git_publish()` (success *and* failure, with `pushed`/`pr_url`), `/api/snapshot` (both outcomes), and `_fail()` (every unhandled API error + traceback tail).
- `GET /api/logs` returns events newest-first — reachable from the browser/curl, no need to shell into the sandbox.
- Lives at `.super-coder/logs/webapp.log` (gitignored). Best-effort: logging never breaks the request it records.

**gh-auth fallback** — `_gh_token()` now falls back to `gh auth token` when no `SC_GH_TOKEN`/`GH_TOKEN` is in the env, so a **host-run** server with a web-authed `gh` works with nothing to export. Sandbox path unchanged (forwarded `GH_TOKEN`); gh-absent fails to `""` cleanly.

**Docs** — `docs/publish-and-gh-auth.md`: how snapshot/publish work, the token chain (`SC_GH_TOKEN → GH_TOKEN → host gh login`), web-auth vs PAT setup, the event log, and a troubleshooting table. Linked from the README Review GUI section.

**Tests** — `tests/test_webapp_log.py` (cap-at-20 + keep-newest, one physical line per multi-line event, custom fields, corrupt-line tolerance, best-effort no-raise). Full suite: **113 passed**.

PR #225, opened 2026-07-01
fix(boot): kill the cwd trap — SC_ROOT + sc on PATH + cwd-independent skill guidance

jedbjorn:
## The problem

A recurring, structural trap reported from a dos-arch fork shell (and present in every fork): the engine + its live DBs sit at the **main worktree root**, but a shell is exec'd from its **own worktree**. To run a convenient `./sc …` or `sqlite3 .super-coder/shell_db.db`, a shell would `cd` to the root — and because Bash cwd persists across calls, every *later* bare `git`/`grep` then silently targeted the main tree (a different branch), so the shell's own worktree edits looked vanished.

The old mitigation was per-command vigilance ("re-prefix every grep with `cd <worktree> &&`") — a rule designed to be forgotten under load, so it recurred each session.

## The fix — cwd-independence by construction

**Mechanism** (`run.py`, the single exec chokepoint for every harness, docker *and* no-docker):
- Export `SC_ROOT` (main worktree root) and prepend it to `PATH` before exec'ing the harness. `sc …` now resolves **bare from any cwd**; raw reads can address the engine by `$SC_ROOT`. No `cd` to the root is ever needed.
- `sc`: add `sc sql "<q>"` / `sc map-sql "<q>"` read passthroughs (absolute-path resolved) so no skill example needs a cwd-relative `sqlite3 .super-coder/…`. Writes still route through `sc mem` (triggers/caps).

**Guidance** (so the docs stop re-teaching the trap):
- Swept 20 engine skills (`assets/skills/*`): `./sc` → `sc`; raw sqlite reads → `sc sql`/`sc map-sql`. Prose that merely *names* a DB file is left intact.
- `0001_seed_skills.sql` regenerated from the edited assets.
- `0036_reseed_cwd_sc_paths.sql`: db_map/memory/messaging/spec were last re-pinned by 0028/0031/0032/0014, so a full replay would end on the stale `./sc` body — re-UPSERT current asset content as the last writer (byte-identical to 0001, same pattern as 0031).
- `skills_sc/*` mirror re-rendered from the DB.

One invariant — *never cd; address the engine by path* — instead of a conditional rule that fails under load.

## Scope notes
- `dev_kit` is fork-owned (not an engine asset) and left as-is.
- Installed forks pick this up via `sc update` → `sync_skills`; fresh builds via the migrations.

## Verification
- `sc render-check` ✓ (hermetic mirror match, byte-identical to CI)
- `test_skills_freshness.py` ✓ (DB built from all migrations matches `assets/skills` exactly)
- Full backend suite: **122 passed**
- Smoke-tested `sc sql` / `sc map-sql` resolve the main engine DBs from a worktree cwd with no `cd`.


jedbjorn:
- engine ref: `eadd9dad3246d317cbc138f598bb9671b96e5c4e` · behind head `b1fa8e9`
- fork/seat: dos-app · Planner shell · sandbox

**Ran / followed:** bootstrap skill examples from the launched worktree cwd (`./sc map-sql …`, then `./sc mem which`).
**Expected:** the documented CLI wrapper to resolve for first-run orientation.
**Actual:** `/bin/bash: ./sc: No such file or directory`; bare `sc` is also absent from PATH in this launched shell.
**Workaround:** invoke the main-root wrapper by absolute path while retaining the worktree cwd. The running shell cannot safely self-update; noting the stale engine ref for triage.

PR #485, opened 2026-07-22
fix(session-control): always-fresh enter + support-latest adapter probes

### LINKED ISSUE #483: [kcsos] Codex adapter: controlled launch rejects Codex 0.145.0
- engine ref: b151e724d2b0a9f5222ea52b4c56b1e839dc2e24 · behind head 94ee48a; confirmed the same 0.144-only gate remains at that HEAD
- fork/seat: kcsos · Admin launching planner PLN2 through host `make dos-e` into the sandbox

**Ran / followed:** `make dos-e`, selected Planning-02 and harness Codex. Installed CLI is `codex-cli 0.145.0`.

**Expected:** The controlled Codex launcher either supports the installed current CLI after protocol validation, or fails before creating/poisoning a session binding with an actionable compatibility message.

**Actual:** `.super-coder/adapters/codex/codex_rpc.py` hard-codes `SUPPORTED_MINOR = 144`, so `probe_codex()` reports `create=false` for 0.145.0 even though the app-server and resume capability probes run. `codex-session.py` then exits with `Codex 0.145.0 is not validated for app-server session control (expected 0.144.x)`. The engine-bundled/current CLI therefore cannot launch a planner through Codex. Upstream HEAD 94ee48a still has the same gate.

**Workaround:** Downgrade Codex to 0.144.x or choose another harness; neither was applied. Please validate 0.145.x and update the supported-version contract/tests.

### LINKED ISSUE #484: [kcsos] session control: failed Codex bootstrap poisons the next enter
- engine ref: b151e724d2b0a9f5222ea52b4c56b1e839dc2e24 · behind head 94ee48a; confirmed the same binding reuse path remains at that HEAD
- fork/seat: kcsos · Admin launching planner PLN2 through host `make dos-e` into the sandbox

**Ran / followed:** Re-ran `make dos-e`, selected Planning-02 and Codex after the initial controlled Codex launch failed. The initial trigger is #483, but this report is specifically about relaunch recovery and error preservation.

**Expected:** A failed controlled bootstrap must not make ordinary re-entry crash. The next enter should either reject the errored binding before spawning with the documented `session retry` or `session release` remedy, or explicitly recover it through `error -> starting` before claiming a foreground lease. The original adapter error should remain visible.

**Actual:** The second launch prints the normal boot summary, then fails:

```
→ session binding: 1 · codex=pending
→ supervise python3 .../adapters/codex/codex-session.py
session launch refused: invalid session binding transition: error -> foreground
make: *** [.super-coder/aliases.mk:39: dos-enter] Error 1
```

Code path: `binding_for_enter()` only selects `managed=1`, so it misses the errored binding created before native-session registration. `open_session()` reuses the still-unused active archive. `ensure_binding()` finds the existing row by `archive_id` and returns it unchanged even when `state=error`. `claim_lease()` then attempts the forbidden direct transition to `foreground`. This masks the useful first-launch error and leaves `dos-e` wedged. Upstream HEAD 94ee48a retains all four conditions.

**Workaround:** `./sc session retry PLN2` can move the row to `starting`, but #483 makes the controlled child fail again; releasing the binding and choosing another harness is the only practical bypass. No workaround was applied.

Suggested regression: create an unmanaged binding, claim/release it with no native session so it reaches `error`, then run the ordinary enter binding-selection/archive-reuse path and assert it produces an actionable recovery error before spawn, never `error -> foreground`.

jedbjorn:
## Root cause

Two stacked session-control defects, plus the RST-C launch crash:

1. **#483 — version pins reject current CLIs.** `probe_codex()` hard-coded `SUPPORTED_MINOR = 144`, so codex-cli 0.145.0 got `create=false` and the controlled launcher exited "not validated". Same pin shape in the claude (`TESTED_VERSIONS = {(2,1)}`) and kimi (`TESTED_SERVER_VERSIONS = {(0,27),(0,28)}`) probes.
2. **#484 + RST-C — enter resumed the managed binding's native thread**, so any stale/poisoned binding wedged or crashed `make dos-e`: an errored binding rode the unused-stub archive reuse into `claim_lease` (`error -> foreground` invalid transition), and a binding whose rollout no longer exists crashed the codex TUI bootstrap (`thread/resume: no rollout found for thread id`).

## Fix

**Interactive enter always starts a fresh session.** `binding_for_enter` → `supersede_for_enter`: the shell's managed binding (or one parked on the active archive) is released — wake jobs cancelled, `last_error` preserved for audit — a fresh archive is opened (`force_new`), and a fresh binding is created with the managed flag carried over. The launcher then opens a NEW provider conversation instead of resuming. A live/orphaned owner still refuses the enter (one shell, one session).

**Headless sprint orchestration is untouched**: the dispatcher's `--session-binding` resume/deliver path, lease fencing, and `./sc session manage|release|retry` all behave as before. Wake jobs rebuild from unread messages on the new managed binding, and the dispatcher safely idles until the new native id registers.

**Version policy: floor, not pin.** codex >= 0.144, claude >= 2.1, kimi >= 0.27 — newer CLIs are accepted on the existing `--help` feature probes (the real capability gate); older/unparsable versions still fail closed. Trade-off, stated: protocol drift in a future CLI now fails loudly at RPC time instead of refusing at boot — that is the intended `support Latest` behavior.

## Testing

- New/rewritten: `SupersedeForEnterTest` (dormant/error/live-owner/unmanaged/no-binding + wake-job cancellation + managed rebind), `EnterMainFlowTest` (supersede-not-refuse at `main()`, RENDER_ONLY never mutates), probe floor tests for all three adapters (newer accepted, older closed).
- Targeted: 104 passed. Full `./sc test`: 654 passed; the 3 `test_vm_bake.py` failures are the documented sandbox-environment baseline (they pass with `SC_SANDBOX` unset; unrelated to this change). Ruff clean. Real-CLI probe smoke: codex 0.144.6 validates, 0.145.0 now accepted by the floor.

## Follow-ups (not in this diff)

- Spec #20 body (DB-owned) still documents "managed bindings resume by default" + SC-466 retry-first — needs a `sc mem doc` update or PLN1 pass; conformance render follows the DB.
- `./sc update-harnesses` on forks pulls the fixed probes; kcsos/RST-C recover via `./sc update` then `make dos-e` (the errored binding is superseded automatically).

Closes #483
Closes #484

### DIFF
--- a/.super-coder/adapters/claude/claude_cli.py
+++ b/.super-coder/adapters/claude/claude_cli.py
@@ -8,7 +8,10 @@ import subprocess
 from collections.abc import Callable
 
 
-TESTED_VERSIONS = frozenset({(2, 1)})
+# Validated version floor — newer CLIs are accepted on feature-probe evidence
+# (the flag detection below), older ones fail closed. Support-latest policy:
+# a version bump alone never disables session control.
+MIN_TESTED_VERSION = (2, 1)
 
 
 def _run(command: list[str]) -> subprocess.CompletedProcess[str]:
@@ -42,7 +45,7 @@ def probe_claude(
     match = re.search(r"(?<!\d)(\d+)\.(\d+)\.(\d+)(?!\d)", version_text)
     version = tuple(map(int, match.groups())) if match else None
     help_text = _output(help_result)
-    known = bool(version and version[:2] in TESTED_VERSIONS)
+    known = bool(version and version[:2] >= MIN_TESTED_VERSION)
     create = (
         known
         and help_result.returncode == 0
--- a/.super-coder/adapters/codex/codex-session.py
+++ b/.super-coder/adapters/codex/codex-session.py
@@ -19,7 +19,9 @@ sys.path.insert(0, str(ENGINE / "scripts"))
 
 import db_driver  # type: ignore[import-not-found]  # noqa: E402
 import session_supervisor  # type: ignore[import-not-found]  # noqa: E402
-from codex_rpc import AppServerClient, probe_codex  # noqa: E402
+from codex_rpc import (  # noqa: E402
+    MIN_SUPPORTED_MAJOR, MIN_SUPPORTED_MINOR, AppServerClient, probe_codex,
+)
 
 
 def socket_path(binding_id: int) -> Path:
@@ -56,7 +58,8 @@ def main() -> int:
     if not probe["create"]:
         raise SystemExit(
             f"Codex {probe.get('cli_version') or 'unknown'} is not validated for "
-            f"app-server session control (expected 0.144.x)"
+            f"app-server session control "
+            f"(requires codex-cli >= {MIN_SUPPORTED_MAJOR}.{MIN_SUPPORTED_MINOR})"
         )
 
     RUN_DIR.mkdir(parents=True, exist_ok=True, mode=0o700)
--- a/.super-coder/adapters/codex/codex_rpc.py
+++ b/.super-coder/adapters/codex/codex_rpc.py
@@ -27,12 +27,16 @@ class CodexRpcError(RuntimeError):
     """The app-server returned a JSON-RPC error."""
 
 
-SUPPORTED_MAJOR = 0
-SUPPORTED_MINOR = 144
+# Validated version floor. Newer CLIs are accepted on feature-probe evidence
+# (the --help detection below is the real capability gate); older ones fail
+# closed. Support-latest policy: a version bump alone must never wedge a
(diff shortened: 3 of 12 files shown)


jedbjorn:
QA repair pushed at faf2ea3: enter supersession now holds one SQLite write lock across ownership reconciliation and release, refuses the dispatching-before-lease window, and adds two deterministic regressions. Local verification: 43 supervisor/dispatcher tests passed; Ruff clean; full suite 656 passed with only the documented 3 sandbox VM-bake posture failures; VM-bake passes 6/6 with SC_SANDBOX unset.

PR #495, opened 2026-07-22
revert: strip docs-skill single-writer note + reseed 0077

jedbjorn:
Per PLN1's ruling (sprint 25, task #357): the single-writer constraint is already durably recorded in decision #20 + roadmap #21, so documenting it in the `docs` skill text was redundant and not worth the three-artifact engine-skill commit (asset + reseed migration + mirror).

This strips from what #494 merged:
- the note in `.super-coder/assets/skills/docs/SKILL.md`
- reseed migration `0077_reseed_docs_single_writer.sql`
- the re-rendered `skills_sc/docs.md` mirror

**Kept:** the r3 lock-scoping revert and the `_CONTENT_WRITE_LOCK` code comment (the real value of #494).

Safe to delete 0077: the migrate ledger never references applied-but-absent files, and any DB that applied 0077 heals back from the reverted asset on the next engine-skill sync.

Local: render-check hermetic ✓, skills-freshness + mem/publish tests 50 ✓.

Co-authored-by: Code-02 (super-coder) <noreply@super-coder.local>


jedbjorn:
Cancelling per PLN1 (planner): #494 is already merged, green, and internally consistent (docs-skill asset + reseed 0077 + re-rendered mirror all agree; render-check passed). The single-writer constraint note in the docs skill is harmless and mildly useful. Reverting working, merged code to make seq 2 'minimal' is net-negative churn — the redirect that motivated this strip was aimed at avoiding the three-artifact WORK before #494 merged, not at removing it after. #494 stands as the final seq-2 state. Thanks for the diligence — this one's on my late redirect, not your work.

PR #506, opened 2026-07-23
docs(sprint_orchestration): faulted worker consumes its task row

jedbjorn:
## What

A worker that faults mid-task — rate-limit cutoff, provider error, dead session — has **already marked its `task` row read**. Workers mark on *start of acting*, not on completion, so the fault leaves a read row and an unfinished unit. `./sc run` on that shell then drains an empty inbox and the worker idles on the default prompt: the re-launch looks like a nudge but does nothing.

Step 4 (Unblock) gains this as its own stall entry:

- confirm the row's state **at runtime** — `sc mem message sent` carries read receipts
- a task row showing read → re-send it (same unit + where the work stopped), *then* boot
- "Link gone quiet" now points at it: a read task row distinguishes a faulted link from a stalled one

## How it's authored

This repo is the engine source, so the skill lands the standard way: asset `SKILL.md` + a reseed migration. `0081` UPSERTs by name, so `skill_id` and existing grants survive. The `skills_sc/` mirror is re-rendered from the hermetic tracked-source build, not from a live DB.

## Verification

`python3 .super-coder/scripts/render_check.py` → `✓ render-check: flat _sc mirror matches the render of the tracked sources` (81 migrations applied clean).

Note: `./sc render-check` resolves the engine at the **main** worktree root, which is currently behind `origin/main` (75 migrations, up to 0077) — so from a linked worktree it checks the wrong tree and reports a stale, unrelated drift on `specs_sc/sprint-planner-session-control.md`. Ran the worktree-local checker directly instead. Flagging separately; not touched here.

## Trade-off

Added a distinct bullet rather than a clause on "Link gone quiet" — the two stalls have opposite remedies (re-task vs. plain boot), and folding them together is what makes the no-op re-launch easy to reach for.

### DIFF
--- a/.super-coder/assets/skills/sprint_orchestration/SKILL.md
+++ b/.super-coder/assets/skills/sprint_orchestration/SKILL.md
@@ -291,10 +291,21 @@ Stalls and the moves:
   Low, reviewer says Medium) -> rule by message immediately — a chain
   waiting on a classification argument is pure loss. Dispute about what
   the unit *should do* -> FnB.
+- **Worker faulted mid-task** (rate-limit cutoff, provider error, session
+  died): its `task` row is already consumed — a worker marks the row read
+  when it starts acting, so a fault leaves a read row and an unfinished
+  unit. Re-launching alone drains an empty inbox and the worker idles on
+  the default prompt. **Confirm the row's state at runtime before you
+  boot** — `sc mem message sent` carries read receipts; a task row showing
+  read means re-send it (same unit, plus where the work stopped and what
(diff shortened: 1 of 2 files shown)


jedbjorn:
Temporarily reverted by #550 to restore green main during the active sprint. Recovery source is preserved at head ba468739a6f4b42aeedc7257d895d0427328dd6c on branch docs/sprint-orch-worker-fault. Resubmit after the sprint from fresh main with a newly numbered trailing reseed generated from the final combined asset, including wake operations.

PR #510, opened 2026-07-23
docs(skills): anticipated user activity — soft-vocabulary posture section for specs

jedbjorn:
## What

Specs now carry a required `## Anticipated User Activity` section — the feature's posture statement: who is expected to touch it, where it can be reached, whose data it holds, and what it does not intend to allow. Soft vocabulary, hard invariants.

- **docs skill** gains the authoring requirement: the section shape (Vocabulary / Expected Activity / Reach / Data Tenancy / Beyond Intention), a shared role roster, the soft-language map, and a one-line escape for internal-only features.
- **spec skill** gains the consumption: Analyze reads the section as governing intent, Verification checks the build against it, plus a stance bullet.
- **0081 reseed migration** UPSERTs both skills (skill_id + grants preserved) so it reaches installed forks on `./sc update` and fresh builds alike. Generated with `seed_skills.parse_skill`/`sql_str` so the inline body matches what a 0001 regen would carry.
- `skills_sc/{docs,spec}.md` mirrors re-rendered hermetically (schema + migrations + content.sql); `render-check` passes.

## The roster

Valid Privileged User · Valid User · Visitor · Future Potential User · **System** (daemons/jobs/watchers) · **Shell** (AI agent shells acting through granted tools) · Unexpected Participant.

## Trade-off, stated

The vocabulary is deliberately soft (no *threat model*, *attack surface*, *adversary*, *exploit*, …) so specs don't trip model content filters — the enforceability lives in precise, code-checkable statements ("a Valid User only ever sees rows tied to their own account") rather than in naming categories of harm. Reviews and the Verification task test the build against the section, so the intention stays binding without the vocabulary.

Older specs predate the section; the spec skill explicitly says absence there is not a blocker.

## Verification

- Migration applies through the full 81-migration chain in a scratch DB; both skills come out with the new section.
- `render_check.py` from this worktree: ✓ mirror matches tracked sources.
- `pytest tests/test_skills_freshness.py tests/test_skill_retire.py tests/test_local_skill_persistence.py` — 34 passed (run in a fresh worktree venv; the main checkout's venv pytest is broken — missing `_pytest` module, flagging separately).

Co-Authored-By: Code-04 (super-coder) <noreply@super-coder.local>

### DIFF
--- a/.super-coder/assets/skills/spec/SKILL.md
+++ b/.super-coder/assets/skills/spec/SKILL.md
@@ -53,6 +53,12 @@ Surface all three before any planning or code:
   slice.
 - No stated done-condition in the spec -> that is the first unclear item.
 
+### Anticipated User Activity
+The spec's `## Anticipated User Activity` section is governing intent: its
+roles, reach, and tenancy invariants shape the plan — access and tenancy
+checks are planned tasks, not afterthoughts. Older specs predate the section;
+absence there is not a blocker.
+
 ### Unclear items
 Anything you cannot act on without guessing:
 - Ambiguous between two interpretations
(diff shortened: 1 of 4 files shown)


jedbjorn:
Temporarily reverted by #550 to restore green main during the active sprint. Recovery source is preserved at head a104de288db4f188a77275e62f31adae872f291c on branch docs/skills-anticipated-user-activity. Resubmit after the sprint from fresh main with newly numbered reseed migrations and hermetic renders.

PR #619, opened 2026-07-26
feat(sprint): the assignment change notice — read off the record (S59 U7)

jedbjorn:
Sprint 59 unit U7 · spec doc 58 "Assignment change notice" · feature 27. Depends on U1, merged as 02d5977.

## What it does

A board write that moves a role now emits one message to each of the shells the record names: the shell newly named, the shell it replaced, and the counterpart role on that unit. The notify set is read **off the record** — no subscription, no reader log, no read tracking — because the audit that retired feature #29 found that in every demonstrated case those are exactly the shells who need telling. The sprint-52 incident is the case: two devs worked for 12 and 27 minutes against a reviewer column that had changed under them.

**No migration.** 0100 stays free. `shell_messages` plus `kind`/`sprint_doc_id` from 0059/0078 is everything this needed.

## The premise correction, first

U1 reported "one write choke point so U7 hooks a single function", and the assignment quoted it back to me. **It is not literally true.** `sprint_units` has exactly **two** writers — `_add_sprint_unit` and `_patch_sprint_unit` — verified against `origin/main` (nothing else in the engine writes the table; `snapshot.py` only names it for the dump). The useful half survives: the writer set is closed and verified, and both writers already sit inside a transaction behind `Idempotency-Key`. So this is one emitter called from two sites, inside each write's own transaction and before its commit. A board write that rolls back tells nobody it happened, and a replayed request returns its cached response without reaching the emitter — which is why there is no `dedupe_key` here.

## Trade-offs taken, all reported pre-build and ruled on

- **`kind='shell'`, deliberately.** `task`/`result`/`pr_event` are exactly `interface_wake.ELIGIBLE_KINDS`, so the natural-reading choice would let a planner's board edit turn into a **boot** of whichever party is a bound planner. This is a notice, not work. Sprint scope is still stamped so the rows filter with the sprint's other traffic.
- **The budget is an invariant, not a count.** The spec said "at most three rows per assignment change"; a PATCH may move both roles in one request, and the union is then four shells. Ruled: the three-party rule is **per changed role**, deduped by recipient, and the invariant to build against is **no shell receives more than one row for one write**. "At most three" was a proxy that holds for a single-role change and nothing else. The rejected alternative — refusing a two-role PATCH the way `state` already refuses to ride along — does not reduce noise, it splits it.
- **Sender chain**, because `from_shell_id` is `NOT NULL` and the operator token carries no shell: `actor.shell_id` → the sprint's bound planner via U1's existing `_board_writer` → the recipient itself, which is `pr_poller.py`'s established convention for daemon-emitted rows. A sender resolving to NULL would turn every operator board edit into a 500.
- **One body**, delivered to each party, rather than three per-role phrasings of one fact.
- **The writing shell is excluded from its own notice.** Can only take the count below the ceiling, never above.
- **No terminal-state gate.** The spec conditions this on the sprint being ACTIVE and nothing else, and no terminal-set constant exists on `main` yet (U4 introduces it), so there is nothing to import and minting a third statement of it would be the drift this spec keeps closing. The ACTIVE gate is on the **notice**, never on the record — a planner can still correct a closed sprint's board, in silence.

## Verification — mutations run, not reported

Spec's row: *assignment change notifies exactly three | reassign a reviewer | notify by any broader rule → extra rows, red.* The test board carries a **second unit** whose shells must stay silent; a single-unit fixture cannot tell "notify the record's parties" from "notify the board". The primary test asserts the recipient **set**, not a count — "three rows" is satisfied just as happily by the wrong three.

All round trips under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` cleared, and the **baseline re-asserted green after every revert** (a same-size same-second revert leaves stale bytecode running; that cost me two false readings in sprint 38).

| # | Mutation | Result |
|---|---|---|
| M1 | notify every shell on the BOARD, not the record | 12 red, incl. `test_reviewer_reassignment_notifies_exactly_three` |
| M2 | drop the actually-moved guard | 3 red (field edit / state move / re-assert now emit) |
| M3 | drop the counterpart from the set | 5 red, incl. the primary |
| M4 | drop the ACTIVE gate | 1 red (`..._closed_sprint_emits_nothing...`) |
| M5 | `kind='task'` — make the notice wake-eligible | 1 red (`test_the_notice_can_never_boot_a_shell`) |
| M6 | drop the self-addressed sender fallback | 17 red (NULL sender → 500) |
| M7 | take the counterpart from the REQUEST, not the row | 4 red — **but not the primary**, see below |
| M8 | emit per changed role instead of per recipient | 6 red, incl. the invariant test |

**M7 is reported against myself.** Passing the request's role subset instead of the merged row leaves the primary test **green** by coincidence: the omitted role then reads as changed-to-None, which puts the same three shells in the set by a different route. Four other tests catch it, so the property is pinned by the suite — but it is *not* pinned by the test that names it. That is the "adjacent property left free" shape, and a reviewer should decide whether it wants a test that fails for the right reason.

**Known limit, stated rather than papered over:** the rollback case is covered by 403 / 404 / 409, all of which return *before* the write. Those prove the emitter is not reached on a rejected request; none of them is a true mid-transaction rollback.

## Suite

`tests/test_sprint_assignment_notice.py` — 22 tests, green. `tests/test_sprint_board_record.py` (U1) — 36, green, no regression.

Full suite via `./sc job`: **1696 passed, 3 failed**. The 3 are `tests/test_harness_epoch.py::ScHarnessCommands` and are **pre-existing and unrelated** — verified by running that file on the main checkout at `d31161a`, where the same 3 fail identically. They assert a stubbed `9.9.9 (Claude Code)` while the real host CLI versions (`claude 2.1.220`, …) leak through the fixture. Not my diff, not patched here.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/.super-coder/api/interface_routes.py
+++ b/.super-coder/api/interface_routes.py
@@ -2407,6 +2407,120 @@ def _unit_projection(con, unit_id: int) -> dict:
     return unit
 
 
+# ------------------------------------------------- assignment change notice
+
+# The two role columns, in one order, derived from the single definition
+# above — the counterpart lookup below indexes into this.
+_ROLE_COLS = tuple(_UNIT_ROLES.values())
+
+
+def _counterpart(col: str) -> str:
+    """The other role on the unit. A unit has exactly two roles — the pair IS
+    the unit — so "the counterpart" is total and needs no default."""
+    return _ROLE_COLS[1 - _ROLE_COLS.index(col)]
+
+
+def _assignment_parties(before: dict, after: dict) -> set:
+    """The shells one board write must tell — the spec's rule and nothing
+    wider (doc 58 "Assignment change notice"): for each role that ACTUALLY
+    moved, the shell newly named, the shell it replaced, and the counterpart
+    role on that unit as the board now reads.
+
+    The set is read OFF THE RECORD. There is no subscription and no reader
+    log because the audit that retired feature #29 found the shells who need
+    telling are exactly the shells the record names — so any rule broader
+    than "these columns" is the defect this emitter exists to avoid, not a
+    generalisation of it.
+
+    A role re-asserted to the same shell has not moved and tells nobody: the
+    sprint-52 incident was a column that CHANGED under two workers, and a
+    planner re-typing the value it already holds is not that.
+    """
+    parties = set()
+    for col in _ROLE_COLS:
+        if before.get(col) == after.get(col):
+            continue
+        parties.update({before.get(col), after.get(col),
+                        after.get(_counterpart(col))})
+    parties.discard(None)
+    return parties
+
+
+def _role_phrase(con, before: dict, after: dict) -> str:
+    named = []
+    for role, col in _UNIT_ROLES.items():
+        if before.get(col) != after.get(col):
+            named.append(f"{role} {_shortname(con, before.get(col))} -> "
+                         f"{_shortname(con, after.get(col))}")
+    roster = ", ".join(f"{role} {_shortname(con, after.get(col))}"
+                       for role, col in _UNIT_ROLES.items())
+    return f"{', '.join(named)}. Now: {roster}."
+
+
+def _shortname(con, shell_id) -> str:
+    if shell_id is None:
+        return "unassigned"
+    row = con.execute("SELECT shortname FROM shells WHERE shell_id=?",
+                      (shell_id,)).fetchone()
+    return row[0] if row is not None else f"shell {shell_id}"
+
+
+def _emit_assignment_notice(con, actor, doc_id: int, seq: str,
+                            before: dict, after: dict) -> int:
+    """The remainder of retired feature #29 (spec doc 58, decision #74): at
+    most three rows per assignment change, only on change, only while the
+    sprint is ACTIVE. Returns the number of rows written.
+
+    Called from inside the board write's own transaction, before its commit,
+    so a board write that rolls back tells nobody it happened — and the
+    routes' Idempotency-Key discipline means a replayed request returns its
+    cached response without reaching this at all. That is why there is no
+    dedupe_key here: the write it rides is already idempotent.
+
+    ONE BODY, delivered to each party rather than three phrasings of one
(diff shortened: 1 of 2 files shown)


jedbjorn:
**Update at b80f774 — two limits I reported against this PR are now closed by planner ruling.**

**M7 — the naming test now fails for the right reason.** The recipient set alone could not distinguish the wrong route (an omitted role reads as changed-to-None, dragging the counterpart in as a "replaced" shell and landing the same three shells). It cannot reproduce the *body*, which reports the dev as vacated by a write that never touched the dev column. `test_reviewer_reassignment_notifies_exactly_three` now asserts what each party is **told** as well as who is told, and M7 reddens it by name. The separate roster test is folded into it — two tests asserting one notice is the restatement shape this unit exists to close, and that split is what let the naming test be weak.

**The rollback case is now pinned by fault injection, not claimed by construction.** A fault is injected after the emitter has run and before the commit; the test asserts the notices **and** the board move are gone together. A notice surviving a failed board write would tell three shells about a reassignment the board never recorded — belief and its announcement disagreeing, which is the defect class this service exists to detect.

The premise was checked before the test was written: `db_driver.connect` leaves `isolation_level` at the default, so the writes really are one deferred transaction and closing without commit really does roll back. Had it been autocommit, the "same transaction" claim in the description above would have been false and this would have been a defect report against my own implementation instead of a test.

| # | Mutation | Result |
|---|---|---|
| M7 | counterpart from the REQUEST, not the row | now reds the primary test **by name** |
| M9 | move the emit after `con.commit()` | reds the rollback test, alone |

10 round trips total, all under `PYTHONDONTWRITEBYTECODE=1` with caches cleared and the baseline re-asserted green after every revert. 22 tests green; all 7 checks green on b80f774.

The 3 local `test_harness_epoch` failures noted above are **flag_id 208** — they exercise a stubbed host docker path and cannot pass inside the sandbox at all, which is why CI is green on them while a local run is not.

PR #621, opened 2026-07-26
fix(sprint): a failed declare must cost one attempt, and a board belongs on a sprint doc — flags 221, 222, 223 (S59 U1 follow-up)

jedbjorn:
U1's post-merge review pass (#616 @baea56a): 1 Major, 1 Medium and one Low, fixed here. U1 is merged, so this is a follow-up unit rather than a rework. Two one-line production changes; the tests are the unit.

## flag 221 (MAJOR) — `sc sprint unit add` permanently bricked a (sprint, seq)

`cmd_unit_add` keyed idempotency deterministically as `unit-add|{sprint}|{seq}`, and `_idempotent` caches whatever `produce()` returns **including its error statuses**. So:

```
add --dev DEV55 (typo)  -> 422 no_such_shell, AND THE 422 IS CACHED
add --dev DEV5  (fixed) -> 409 idempotency_conflict
any other body          -> 409
replay the typo         -> 422, the only body that still resolves
```

**The trigger and the instructed repair are the same action** — a typo'd shortname on the declare path, and the retry the error message asks for. Permanent, not 24h: `expires_at` is written and indexed but never read, nothing deletes from `interface_idempotency_keys`, and the table is snapshot content, so it survives a rebuild.

`add` now suffixes a `uuid4()` exactly as `set` and `state` already do — it was the one verb that did not.

**The Low closes in the same change.** With a fresh key, a duplicate declare finally reaches the route's own refusal: `409 unit_exists` / "edit it with PATCH rather than declaring it twice" — the answer the migration comment, the route docstring and the CLI all present as *the* answer to a double declare, and which was unreachable on the shipped path because the cached key answered first. The code was right; nothing could get to it.

Trade-off, stated: a genuine duplicate is now caught by the natural key `(sprint_doc_id, seq)` rather than by the idempotency key. That is the weaker guarantee on paper and the correct one here — the key was protecting an effect that had not happened.

## flag 222 (MEDIUM) — a released binding is still an answer

`_board_writer` filtered `released_at IS NULL`, which made the fallback mean "before a binding is armed" **or** "after the planner's session ended" **or** "after the sprint was closed or frozen". Three engine paths release with no operator action at all (`shell_recovery`, `update_discard`, `_close_sprint_wake`). In that window the fence widened from one shell to a *class*: a foreign planner could move a unit to `merged` — terminal, the state that makes the reconciler stop watching it — on a sprint it has no relationship to.

The clause is deleted. The query already ended `ORDER BY binding_id DESC LIMIT 1`, so the fallback is now "the most recent planner ever bound to this sprint" — one shell, the shape the spec's rule has in both of its branches — and flavor stands in only where a sprint has never had a binding.

**Deliberate cost, ruled by the planner:** a new planner taking a sprint over must arm its binding before it can write the board. That is one command, against a foreign planner silently terminating units.

## The test that would have caught it

Nothing drove the CLI's key strategy: `_BoardCase.call` mints a fresh key per call, and `test_each_verb_calls_the_method_and_path_it_claims` spies `_api` with a signature that drops the key argument on the floor. So the new cases patch `urlopen` **underneath** `_api` and run the shipped verbs against the real in-process routes — the key that reaches `_idempotent` is the one the shipped code assembles, header casing and all.

Six new tests (42 total in the file, up from 36):

- a failed declaration can be corrected — the 221 regression, end to end through the verb;
- every mutating verb mints a fresh uuid4-suffixed key under its readable prefix, and reads carry none;
- a redeclaration reaches the route's refusal — asserted on **the sentence the planner is told**, because a body mismatch and a duplicate unit are different mistakes with different repairs;
- a released binding still names the board writer, and the sprint's own planner still writes;
- a released binding does not open the sprint next door (the cross-sprint leg);
- a handover leaves the board with the planner that *took* it.

That last one is not from the review. Once a released binding is part of the answer, *which* released binding answers becomes load-bearing, and every other case has one binding on the sprint and cannot tell `DESC` from `ASC`. No mutation of mine had reached it; it was enumerated rather than waited for.

## Verification

Four mutation round trips, `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` cleared, **baseline re-asserted green after every revert**:

| mutation | reds |
|---|---|
| `add`'s key loses its uuid (the shipped defect) | the 3 CLI-key tests |
| `add` keys on a bare uuid, no readable prefix | the key-shape test |
| restore `AND released_at IS NULL` (the shipped defect) | the 3 fence tests |
| read the bindings oldest-first | the handover test |

Full suite: **1689 passed**, 3 failed — `tests/test_harness_epoch.py`, flag_id 208, the stubbed host docker path that cannot run in this sandbox and is green in CI. No migration; 0100 stays free.

## Also here

One commit folds REV1's non-blocking Low from U7 (#619, merged as 3e74b36): `_emit_assignment_notice`'s docstring still claimed "at most three rows per assignment change" after the A1 ruling replaced that budget — a dual-role PATCH tells four shells. Restated as the invariant it was a proxy for: at most one row per shell per write. Folded here rather than reopening a merged unit, and declared to the planner rather than done silently.

**Not in scope: flag 224** — `_idempotent` caching non-2xx *at all*, the general defect underneath 221, which the reviewer filed separately and which is every caller's problem, not this unit's.

### DIFF
--- a/.super-coder/api/interface_routes.py
+++ b/.super-coder/api/interface_routes.py
@@ -2321,10 +2321,24 @@ def _board_writer(con, sprint_doc_id: int) -> "int | None":
     `documents` carries no author column at all (schema.sql:204), so there is
     nothing to fall back TO. An unbound sprint therefore falls back to flavor
     instead — see _may_write_board.
+
+    RELEASED BINDINGS STILL NAME THE WRITER. Filtering on `released_at IS
+    NULL` made the fallback mean "before a binding is armed" OR "after the
+    planner's session ended" (interface_recovery release_binding
+    'shell_recovery') OR "after the sprint was closed or frozen"
+    (_close_sprint_wake) — three engine paths that release with no operator
+    action at all. In that window the fence widened from ONE shell to a CLASS
+    of shells, and a foreign planner could move a unit of a sprint it has no
+    relationship to into `merged` — the terminal state that makes the
+    reconciler stop watching it. The spec's rule names one shell in both of
+    its branches, so this one does too: the most recent planner ever bound to
+    this sprint. Flavor now stands in only for a sprint that never had a
+    binding. The cost is deliberate — a NEW planner taking a sprint over must
+    arm its binding before it can write.
     """
     row = con.execute(
         "SELECT planner_shell_id FROM sprint_planner_bindings "
-        "WHERE sprint_doc_id=? AND released_at IS NULL "
+        "WHERE sprint_doc_id=? "
         "ORDER BY binding_id DESC LIMIT 1", (sprint_doc_id,)).fetchone()
     return row[0] if row is not None else None
 
--- a/.super-coder/api/interface_routes.py
+++ b/.super-coder/api/interface_routes.py
@@ -2560,7 +2560,15 @@ def _sprint_units(actor, query: dict):
     if doc_id is not None:
         sql += " WHERE sprint_doc_id=?"
         params.append(doc_id)
-    sql += " ORDER BY sprint_doc_id, seq"
+    # LENGTH BEFORE VALUE, because `seq` is TEXT and a plain sort is
+    # lexicographic: U1, U2, U3, U10, U11 reads back as U1, U10, U11, U2, U3.
+    # The board is a WORK ORDER — a planner reads it top to bottom and a
+    # reader that scrambles at U10 misreports what runs next. Sorting by
+    # length first makes same-prefix numbering natural (every U0-U9 before
+    # every U10-U99) without inventing a numeric column the planner would
+    # then have to keep in sync. Seqs of unlike shape (U-H beside U9) still
+    # order by length then value — deterministic, and no worse than today.
+    sql += " ORDER BY sprint_doc_id, LENGTH(seq), seq"
     con = _db()
     try:
         units = [_unit_projection(con, r[0])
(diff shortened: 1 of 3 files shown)


jedbjorn:
**Scope completed** — the planner ruled flag 223 in and forwarded the two Lows I could not reach (#2303). Three more commits.

### flag 223 (MEDIUM) — a board could be minted on any document

`sprint_doc_id` was validated as "a document that exists", and the FK carries no kind constraint. A sprint doc and its spec are consecutive ids — **59 and 58 for this very sprint** — so `--sprint 58` minted a whole phantom board on the spec: rows the reconciler watches, on a document no participant reads. The seq axis of that typo class was already defended (PATCH never creates, POST is not an upsert); the doc axis was not.

Uses the engine's own definition of a sprint document — `pr_poller`'s `kind='doc' AND title LIKE 'SPRINT:%'` — rather than a second one that can drift from it.

**One judgement call, declared:** that predicate also carries `frozen=0`, and this check omits it. Whether a frozen board stays mutable is a live question the reviewer parked for the FnB and the planner explicitly declined to rule on; taking the whole clause would have decided it as a side effect of a typo guard. The frozen case is left exactly as it was, and the test says so rather than pinning it in either direction.

The refusal names the document it hit, because the planner's next move is to retype the id. A document that is absent keeps its own 404 — two mistakes, two repairs.

### Low A — the board read back in the wrong order

`ORDER BY seq` on a TEXT column is lexicographic: `U1, U2, U3, U10, U11` reads back as **U1, U10, U11, U2, U3**, and no test covered ordering at all. The board is a work order read top to bottom. Length before value makes same-prefix numbering natural without inventing a rank column a planner would then have to keep in sync. Sprint 59 stops at U9, so the first sprint to reach ten units would have been the one to find this.

Asserted on the rendered board as well as the API — the render is what a planner actually reads, and a projection can re-sort.

### Low C — `--title none`

Stored the four letters as the title, while every other field spells retraction that way and the CLI's own help says so. A unit cannot be untitled, so refusing is the honest answer: it names which of the two things the planner meant, where storing it produced a board that was wrong and looked deliberate. Refused on both `add` and `set`; `none of the above` is still a title.

### Verification

45 tests (42 → 45). Three more mutation round trips, same protocol — `PYTHONDONTWRITEBYTECODE=1`, caches cleared, baseline re-asserted green after every revert:

| mutation | reds |
|---|---|
| ordering back to lexicographic | the work-order test |
| the sprint-doc predicate accepts any document again | the phantom-board test |
| `--title none` stored again | the title-refusal test |

**Not taken, on the planner's ruling:** authored row order as a stored column (dependency order is planner presentation — better derived from `depends_on` in U4-adjacent work than stored as a fourth thing to keep in sync), the frozen-doc question, and flag 224.

PR #641, opened 2026-07-26
fix(sprint): render reconciler heartbeat

jedbjorn:
Closes flag 302 / S59-U5f.

Summary:
- return the reconcile heartbeat beside the existing watch heartbeat without breaking the flat watch API fields
- render a dedicated reconciler liveness line with the same age > 3 * interval stale rule
- pin a fresh watch plus 3600-second-stale reconcile scenario through the real DB reader and CLI renderer

Verification:
- 122 tests green: test_sprint_eventing, test_pr_poller, test_reconciler_delivery
- watch-only reader mutant: new scenario red
- dropped reconciler rendering mutant: new scenario red

Trade-off: retain the existing watch fields at daemon top level and nest only reconcile, preserving current clients while adding the smallest read-side surface.

### DIFF
--- a/.super-coder/api/server.py
+++ b/.super-coder/api/server.py
@@ -2733,9 +2733,11 @@ class Handler(BaseHTTPRequestHandler):
     def _daemon_state(self, con) -> "dict | None":
         """Poller + reconciler liveness (#359): heartbeat rows → age + verdict.
         Stale = beat older than 3× the daemon's own interval (one slow gh
-        call + the sleep fit comfortably inside). None = never run (or a
-        pre-0068 DB with no heartbeat table yet) — the client renders both
-        None and stale as "watches are NOT being polled"."""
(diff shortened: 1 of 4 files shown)


jedbjorn:
**PLANNER HOLD — do not merge. Converted to draft deliberately.**

The reviewer (REV1) placed a hold on the EVIDENCE behind its REVIEW-CLEAN verdict after sending it — an instrument fault on its side (concurrent pytest runs sharing one worktree, caused by checking for stray processes with `ps`, which does not exist in this container). The verdict is not withdrawn and is expected to stand; what is unverified is the completeness of the mutation red sets.

A strictly sequential re-verification is in flight. This PR returns to ready when the reviewer confirms, and the merge is re-authorised by the planner (PLN1) — not before.

Nothing is wrong with the code or with this branch. Draft status is enforcing a review gate, not signalling a defect.

PR #730, opened 2026-07-29
fix(conductor): persist failed refusal escalations

### LINKED ISSUE #724: [dos-app] conductor: re-task on merged unit returns HTTP 500
- engine ref: 7ba981a1acc1c70ed85ac5fe4f399b601b10a8a8 · behind head d1ef41d
- fork/seat: dos-app · Conductor · sandbox

**Ran / followed:** `sc directives act 8` for a pending `planner:re-task` targeting U1 after post-merge conformance found a Major defect. U1 was `merged` and still assigned to DEV1.

**Expected:** The mechanical re-task transition reopens U1 into a live developer state and launches DEV1, or refuses the directive with a domain response that explains the unsupported transition.

**Actual:** `POST /api/directives/8/act → HTTP 500: Internal Server Error`. Container log: `ConductorLaunchError(launch pid 6515 exited 1)`. Launch log: `sc run: shell DEV1 has no live dev assignment in sprint document 2 unit U1`. The directive remains pending and U1 remains merged.

**Workaround:** None. I did not bypass the transition or mutate the board directly. The active sprint prevents updating the stale engine, so the staleness caveat cannot be cleared in place.

jedbjorn:
Closes #724.

When a directive is correctly refused but the follow-up Planner slot cannot launch (for example because the sprint was frozen/closed), `ConductorLaunchError` escaped the refusal handler. The connection rollback left the directive pending and permanently poisoned the queue.

This makes refusal escalation best-effort: the directive and refusal event commit with the launch error recorded.

Verification:
- `pytest -q tests/test_conductor_runtime.py` — 19 passed, 23 subtests
- `pytest -q tests/test_conductor_contracts.py tests/test_boot_sprint_directive.py` — 30 passed
- `./sc render-check`
- `./sc verify`
- `git diff --check`

### DIFF
--- a/.super-coder/scripts/conductor_runtime.py
+++ b/.super-coder/scripts/conductor_runtime.py
@@ -1043,7 +1043,7 @@ def _act_locked(
                 model=model,
             )
             escalation = {"command": command, "pid": launcher(command)}
-        except DirectiveRefused as escalation_error:
+        except (DirectiveRefused, ConductorLaunchError) as escalation_error:
             escalation = {"error": str(escalation_error)}
         _trail(
             con,
--- a/.super-coder/scripts/conductor_runtime.py
+++ b/.super-coder/scripts/conductor_runtime.py
@@ -1006,7 +1006,7 @@ def _act_locked(
             "launches": launches,
             "pids": pids,
         }
-    except DirectiveRefused as exc:
+    except (DirectiveRefused, ConductorLaunchError) as exc:
         con.execute("ROLLBACK TO conductor_act")
         con.execute("RELEASE conductor_act")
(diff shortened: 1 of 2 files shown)


jedbjorn:
Extended the fix after downstream Sprint 4 conformance reproduced the same atomicity gap on the normal execution path: a Planner answer targeting a terminal unit attempted to launch REV1, raised ConductorLaunchError, returned HTTP 500, and left the directive pending. New commit catches worker launch failures at the transaction boundary, rolls back the board transition, durably refuses the directive, and escalates to Planner. Regression: 33 passed + 23 subtests; render-check and diff-check green.

Issue #960, opened 2026-08-02
tests: local skill command fixture leaks real worktree state


jedbjorn:
Where: subfloor test suite on a host seat with an existing dev1 worktree.\n\nRan: ./sc test (also reproducible with pytest test_local_skill_persistence.py::SkillCommandTest::test_grant_revoke_roundtrip_by_shortname_and_id).\n\nExpected: the synthetic minimal skill DB remains self-contained.\n\nActual: skill projection discovers the real dev1 checkout, then queries the synthetic DB for resolved_shell_skills and fails with sqlite3.OperationalError: no such table: resolved_shell_skills. The same test passes from a detached base worktree whose repo root has no dev1 checkout.\n\nWorkaround: none applied; U1 verification uses scoped relevant suites. The fixture should isolate projection roots or include the view it now requires.

Issue #959, opened 2026-08-02
tests: restricted restart fixture creates no pre-restart DB backup


jedbjorn:
Where: subfloor test suite, reproduced on untouched base 3a89a5d.\n\nRan: pytest tests/test_supervision_sc.py::RestrictedLaunchTests::test_restart_bounces_systemd_broker_and_reports_full_inventory.\n\nExpected: exactly one shell_db.prerestart.*.db under the fixture backup directory.\n\nActual: restart exits 0 but the glob returns zero backup files.\n\nWorkaround: none; this is outside Sprint 87 U1 chat-registry scope.

jedbjorn:
Reproduced during Sprint 87 U3 red repair on 2026-08-03. The exact test fails both on PR #963 head `7fc7290` and unchanged base/main `89e604e`, each with `len(backup_files) == 0` after a successful restart. The rest of the PR #963 CI-equivalent run passed (1670 passed, 1 skipped, 867 subtests); this confirms the failure is pre-existing and unrelated to the conversation-reaper diff.

jedbjorn:
Reproduced again during Sprint 87 U6 review-fix verification on 2026-08-03. The exact test fails identically on PR #976 head `a51f0cc` and clean current main `2537e49`: restart exits 0, but `shell_db.prerestart.*.db` count is 0. The PR full suite otherwise reported 1695 passed / 1 skipped, and the focused U6 watcher + removal-manifest gates pass. No workaround or unrelated fix applied in U6.

jedbjorn:
Reproduced independently on 2026-08-03 from PR #975's linked worktree with: `.venv/bin/pytest -q tests/test_supervision_sc.py::RestrictedLaunchTests::test_restart_bounces_systemd_broker_and_reports_full_inventory`. The test fails in isolation: restart exits 0, but the expected `shell_db.prerestart.*.db` count is 0. PR #975 does not modify `sc`, `db_backup.py`, or `test_supervision_sc.py`; its focused and adjacent suites pass. This is therefore an existing baseline failure, not the U4 review fix.

Issue #261, opened 2026-07-02
[rst-c] vm-broker: /exec returns raw UnicodeDecodeError when guest stdout is not valid UTF-8


jedbjorn:
- engine ref: 561b247f84834d1a628c2660aa3229742c5a7607
- fork/seat: rst-c · dev shell · sandbox

**Ran / followed:** windows_devkit skill exec loop — `curl --unix-socket $(sc vm-broker-sock) http://vm/exec -d '{"command":"powershell ... Get-Content <file> -Raw"}'` where the file is UTF-16/BOM (a Revit `.addin` manifest).

**Expected:** `{ok, exit, stdout, stderr}` with the guest output (lossy-decoded or bytes-preserving), per the skill's exec contract.

**Actual:**
```
{"ok": false, "error": "UnicodeDecodeError: 'utf-8' codec can't decode byte 0x83 in position 81: invalid start byte"}
```
The whole exec fails — no exit code, no partial output — whenever the guest prints any non-UTF-8 byte (UTF-16 files, OEM-codepage console output, binary spills).

**Workaround:** have the guest base64-encode file/console output (`[Convert]::ToBase64String(...)` or `-replace '[^\x00-\x7F]'`) and decode broker-side. Works, but every fork driving a Windows VM will trip on this — Windows tooling emits non-UTF-8 constantly. Suggest decoding with `errors="replace"` (or returning stdout_b64) in the broker.

PR #961, opened 2026-08-02
feat(conversations): add active chat registry

jedbjorn:
## Summary

- add migration 0162 with one durable active chat per shell and dirty-install convergence
- make browser chat replacement close and commit before registering a replacement
- record PID plus Linux process-start ticks atomically with native run start and clear them on finalize
- make broker dispatch, Sprint chat selection, snapshots, and removal fixtures use the registry authority

## Verification

- focused conversation, adapter, migration, broker, and Sprint coverage: 191 passed, 184 subtests passed
- full suite: 1658 passed; 3 unrelated failures reproduced independently
  - #960 covers the two local-skill fixture failures caused by leaking a real worktree
  - #959 covers the existing restricted-restart backup fixture failure
- diff check, Python compilation, and focused Ruff check pass

## Scope note

The browser creation endpoint implements the required two-commit close/create boundary. The legacy Sprint helper still participates in its caller-owned domain transaction while adopting universal registry arbitration; Sprint 87 U2 owns replacement of the wake producer path.

Sprint 87 · U1 · task #260 · spec doc #89

### DIFF
--- a/.super-coder/api/conversation_routes.py
+++ b/.super-coder/api/conversation_routes.py
@@ -28,6 +28,7 @@ DB_PATH = ENGINE / "shell_db.db"
 SCRIPTS = ENGINE / "scripts"
 sys.path.insert(0, str(SCRIPTS))
 
+import active_chat_registry
 import conversation_broker
 import conversation_events
 import conversation_git_targets
@@ -498,12 +499,8 @@ def _reopen_conversation(con, operator: dict, conversation) -> list[str]:
             "shell is unknown, deleted, or unavailable to this operator",
         )
     _refuse_admin_browser_chat(shell)
-    open_conversations = con.execute(
-        "SELECT conversation_id,state FROM conversations "
-        "WHERE shell_id=? AND state!='closed' AND conversation_scope='normal'",
-        (conversation["shell_id"],),
-    ).fetchall()
-    if not open_conversations:
+    active = active_chat_registry.get(con, int(conversation["shell_id"]))
+    if active is None:
         live_state = _live_shell_session(shell)
         if live_state is not None:
             raise ApiError(
@@ -516,32 +513,31 @@ def _reopen_conversation(con, operator: dict, conversation) -> list[str]:
                     "state": live_state,
                 },
             )
-    running = [
-        row for row in open_conversations
-        if row["state"] in ("queued", "running")
-    ]
-    if running:
+    try:
+        closed = active_chat_registry.close_active(
+            con,
+            int(conversation["shell_id"]),
+        )
+    except active_chat_registry.ActiveChatBusy as exc:
         raise ApiError(
             409,
             "BROWSER_CHAT_BUSY",
-            "the open browser chat has a turn in progress",
-            {"conversation_id": running[0]["conversation_id"]},
-        )
+            "the active chat has a turn in progress",
+            {
+                "conversation_id": (
+                    active.chat_id if active is not None else None
+                )
+            },
+        ) from exc
     auto_closed = []
-    for row in open_conversations:
-        con.execute(
-            "UPDATE conversations SET state='closed',"
-            "closed_at=datetime('now'),last_activity_at=datetime('now'),"
-            "version=version+1 WHERE conversation_id=?",
-            (row["conversation_id"],),
-        )
+    if closed is not None:
         _append_event(
             con,
-            row["conversation_id"],
+            closed.chat_id,
             "conversation.closed",
             {"status": "closed", "reason": "another browser chat reopened"},
         )
-        auto_closed.append(row["conversation_id"])
+        auto_closed.append(closed.chat_id)
     con.execute(
         "UPDATE conversations SET state='idle',closed_at=NULL,"
         "last_activity_at=datetime('now'),version=version+1 "
@@ -553,6 +549,11 @@ def _reopen_conversation(con, operator: dict, conversation) -> list[str]:
         conversation_id,
         "conversation.reopened",
         {"state": "idle"},
+    )
+    active_chat_registry.register(
+        con,
+        int(conversation["shell_id"]),
+        str(conversation_id),
     )
     return auto_closed
 
@@ -754,12 +755,7 @@ def _create_conversation(con, operator: dict, headers, body: dict):
             "shell is unknown, deleted, or unavailable to this operator",
         )
     _refuse_admin_browser_chat(shell)
-    open_conversations = con.execute(
-        "SELECT conversation_id,state FROM conversations "
-        "WHERE shell_id=? AND state!='closed' AND conversation_scope='normal'",
-        (shell_id,),
-    ).fetchall()
-    if not open_conversations:
+    if active_chat_registry.get(con, shell_id) is None:
         live_state = _wait_for_cli_release(shell)
         if live_state is not None:
             raise ApiError(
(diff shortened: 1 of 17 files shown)


jedbjorn (commented):
## S87 U1 adversarial review — task #260 vs spec doc #89 (REV1)

Verdict for the merge gate: **0 Major / 3 Medium / 4 Low**. Medium+ gates merge per sprint protocol; dispositions are PLN1/FnB's call. Contract walked: registry shape, close-then-create fail-loud ordering, same-transaction pid+start_ticks registration, finalize pid-clear, universality. Core ordering and transaction boundaries in `conversation_routes.py` + `conversation_broker.py` check out — the two-commit close/create split, the `ACTIVE_CHAT_CHANGED` re-check, the registration-failure rollback test, and the tolerant `clear_process` on finalize all match the contract. Findings below are where the diff diverges or leaves a hole.

### Medium

**M1 — opencode registers the shared engine-wide server pid as the turn's process identity.**
`conversation_adapters/opencode.py` (`_turn`): `process_ref=str(_SERVER_PROCESS.pid)`. `_SERVER_PROCESS` is a module-global singleton (`opencode serve --port 4096`, `start_new_session=True`) shared by every opencode conversation and living across turns — not "pid of the in-flight turn's process" (spec, registry shape). Two concurrent opencode shells write the *same* pid+start_ticks into two registry rows. Nothing malfunctions inside U1, but this durable identity is exactly what U3's reaper consumes: displace one opencode chat while the others sit idle (pid NULL between turns) and the sweep sees the shared server as an unprotected candidate — the kill ladder then takes down the transport for every opencode shell. Contrast codex, which is correct here: the adapter is built per run (`conversation_broker.py` `_execute`, closed in `finally`), so its app-server pid is genuinely turn-scoped, and its non-numeric fallback records no pid. Suggest opencode record no pid (None, like the codex fallback) until a per-turn process identity exists.

**M2 — migration 0162 strands pending queued work on every chat it closes, bypassing the engine's own cancellation discipline.**
Every runtime close path runs `_cancel_queued_turns` (`conversation_routes.py:573-586`): messages → `cancelled`, outbox rows → `cancelled`. The migration instead walks `queued→idle` (deliberately, to slip past the transition trigger) and closes, leaving `conversation_messages` rows `queued` and `conversation_outbox` rows `pending` forever — and the new `claim_next` registry join guarantees they are permanently unclaimable. A downstream install upgrading with >1 open chat per shell (normal + sprint — legal pre-migration) keeps only the newest and silently loses the rest's pending work: no cancellation, no event, no surfacing. That's a fail-loud violation at the exact boundary this unit's contract is about, and migration integrity / downstream update compatibility is where Sprint-51-style breakage lives. Suggest the migration cancel messages/outbox rows for the chats it closes (mirroring `_cancel_queued_turns`), or at minimum emit events naming what it dropped.

**M3 — sprint-scope exemption retired before U2's delivery rule exists: interim hard-fail against a busy non-linked chat.**
`sprint_participant_chats.py` `create_and_select`: on `ActiveChatBusy`, the new code re-enters only when the active chat is *linked to this participant*; otherwise it raises `SprintConversationError`. The spec's own scenario — a live FnB↔shell chat mid-turn when a sprint New arrives (e.g. planner-bound work while the FnB is talking to the planner) — now burns a wake delivery attempt per retry; three failures auto-pause the sprint. Pre-PR the sprint chat coexisted (exemption); post-U2 the delivery rule re-enters the active chat. This PR retires the exemption before its replacement lands, against the build-order doctrine (retirements last, after replacements proven). Aware the caller-owned-transaction rotation in this helper is PLN1-ratified deferred to U2 — this finding is the *delivery semantics regression window*, not the transaction boundary. May still be ruled deferred-with-U2; flagging so the window is chosen, not accidental.

### Low

**L1 —** the reopen path (`conversation_routes.py` `_reopen_conversation`, called inside the `conversation.message.create` transaction) performs close + reopen + register in one transaction, contradicting `active_chat_registry.py`'s own module contract ("chat rotation must commit `close_active` before beginning the transaction that... registers a replacement") and the spec's literal close-commits-first ordering. The atomic swap actually preserves the invariant (no second chat, rollback is fail-loud), so this is a consistency/doc divergence — but one path or the docstring should yield.

**L2 —** the `ACTIVE_CHAT_CHANGED` race branch (second transaction in `_create_conversation`) has no test; it's the one guard protecting the inter-commit window the two-transaction design opens.

**L3 —** migration 0162's `queued→idle` and `running→error` edge-walks are untested — `test_active_registry_migration_converges_legacy_open_chats` seeds only idle chats, so the trigger-compatibility paths the migration comments call out never execute in CI.

**L4 —** runtime starvation of unlinked open chats is pinned as *silent*: `test_registry_excludes_an_unlinked_legacy_conversation` asserts the queued message just sits `queued` forever. That state is supposed to be unreachable post-migration; per the fail-loud doctrine an impossible state deserves a loud guard (error the message, or log), not quiet permanent starvation. Also nit, same family: in `_reopen_conversation` the busy payload's `active.chat_id if active is not None else None` else-arm is dead — `ActiveChatBusy` cannot raise when `active` is None.

Ruled scope respected: the legacy helper's caller-owned domain transaction (U2/task #261) is not filed; pre-existing failures #959/#960 not filed. Merge decision is FnB-gated — this review is comment-only.

— REV1 · Sprint 87 · U1 · spec doc #89

jedbjorn (commented):
## S87 U1 re-review @ 7cd8de7 (delta from e527861)

**Verdict: review-clean — 0 Major / 0 Medium.** Both Mediums verified resolved at head; no regressions found in the fix commit.

### M1 — opencode shared-server pid as turn identity: resolved
- `conversation_adapters/opencode.py` now records `process_ref=None` for turns; the only remaining `_SERVER_PROCESS` references are server lifecycle management, none flow into turn identity.
- Parity with the codex fallback confirmed at the registry seam: codex's non-subprocess fallback (`app-server:<run_ref>`) resolves through `active_chat_registry.process_identity` to `(None, None)` because it is non-decimal; opencode's explicit `None` takes the same path. The broker's `mark_native_started` then passes NULL identity into `set_process`, satisfying the registry's paired-NULL CHECK. Registry identity is now turn-process-or-NULL, as the U3 reaper contract needs.
- `test_opencode_shared_server_is_not_the_turn_process` pins the behavior (live mocked server, `poll` never consulted), not just the current state.

### M2 — migration 0162 stranding queued work: resolved
- The two new UPDATEs mirror the runtime close path `_cancel_queued_turns` (sprint_participant_chats.py:216) statement-for-statement: messages `accepted/queued/running → cancelled` with `completed_at` set, outbox `pending/claimed → cancelled` with claim fields cleared, and the messages-before-outbox ordering preserved so the message UPDATE can still key off live outbox states.
- Placement is correct: after seed selection, before the state-walk/close, while `_active_shell_chat_seed` still exists; the active chat's queue is excluded via NOT EXISTS.
- The regression test now seeds queued(pending-outbox), running(claimed-outbox), and newest-idle(pending-outbox) chats and asserts the queued→idle→closed and running→error→closed walks, both cancellations (completed_at set, claim_owner cleared), and the surviving active chat's queue intact. This also covers round-1 L3.
- Scope note, no action needed: the sweep additionally cancels stale pending/claimed outbox rows under chats that were already closed before the migration (the runtime only cancels at close time). That is invariant-restoring — such rows were exactly the permanently-unclaimable state M2 targeted.

### Regression check
- Touched test files pass locally at head: 50 tests, 141 subtests.
- Full local suite: 1657 passed; the 5 local failures (local skill persistence ×2, migration CLI help, systemd supervision, sprint help output) are host-environment tests with zero overlap with this diff, and CI is fully green at this head (tests / verify / render-check / CodeQL).

### Standing items
- M3 stands as deferred-with-U2 (PLN1 ruling) — out of scope for this round.
- L1/L2/L4 remain report-only; L3 is now closed by the new migration test.
- Pre-existing round-1-scope observation (report-only, not introduced by the delta): unlike the runtime close path, the migration emits no close events and leaves any leased/starting/running `conversation_runs` rows to lease-expiry recovery.

Not merged, nothing pushed — merge gate stays with the FnB.

Issue #962, opened 2026-08-02
tests: SupervisionFixture inherits SC_ROOT and escapes its fake fork


jedbjorn:
Where: subfloor source checkout, host dev shell.\n\nRan: python3 -m unittest tests.test_supervision_sc.RestrictedLaunchTests.test_restart_bounces_systemd_broker_and_reports_full_inventory -v\n\nExpected: SupervisionFixture restart backs up its temporary restricted-fork shell_db under the fixture HOME.\n\nActual: SupervisionFixture copies os.environ including the launched shell's SC_ROOT. The copied sc dispatcher resolves that inherited root, reports a backup for the live HOME substrate repo, and the fixture assertion finds zero backup files. The isolated test fails consistently with len(backup_files) == 0.\n\nImpact: this test is not isolated when run from a launched shell and can read/mutate the wrong repo substrate.\n\nWorkaround: run only U3-focused suites; proposed fix is for SupervisionFixture to remove SC_ROOT (and other root-routing variables) from its child env before invoking the copied dispatcher.

PR #963, opened 2026-08-02
feat(conversations): reap unlinked process groups

jedbjorn:
## Summary\n- persist run PID, /proc start ticks, process group, and reaper ladder state in migration 0163\n- start a configurable 60s reaper service beside the conversation broker; protect only exact active-registry identity matches and never signal recycled PIDs\n- advance harness interrupt → SIGTERM → SIGKILL across heartbeats, then commit cancelled durable state plus terminal run.interrupted evidence (never unknown)\n- give Claude, Codex, and Kimi turn-owned process groups and shared descendant cleanup; OpenCode's shared server intentionally remains NULL/unowned\n- leave unprotected process-bearing recovery rows to the reaper instead of broker reconciliation\n\n## Verification\n- 131 passed, 174 subtests: reaper, conversation schema/adapters/broker, runtime startup wiring, migration management, Sprint-removal manifest, and sandbox --init coverage\n- focused ruff check + format check for new files\n- focused mypy for conversation_reaper.py\n- git diff --check\n\n## Contract notes\nThe existing conversation run-state vocabulary stores interrupted outcomes as state=cancelled; the terminal semantic record is event_type=run.interrupted with reaper evidence. The reaper never writes unknown. Sandbox launch already uses Docker --init, and starting the reaper from the API runtime keeps /proc inspection inside the harness PID namespace.\n\nKnown unrelated test-isolation defect discovered during verification: #962 (SupervisionFixture inherits SC_ROOT and escapes its fake fork). The relevant sandbox --init supervision test passes.

### DIFF
--- a/.super-coder/api/server.py
+++ b/.super-coder/api/server.py
@@ -4324,14 +4324,17 @@ def main(argv):
             )
             sprint_pr_watcher.start_service(DB_PATH, repo_root=REPO_ROOT)
 
-        await transport.serve(
-            bind,
-            port,
-            dispatch_http,
-            _ws_unavailable,
-            on_started=start_runtime_services,
-            stream_handler=conversation_routes.stream_events,
-        )
+        try:
+            await transport.serve(
+                bind,
+                port,
+                dispatch_http,
+                _ws_unavailable,
+                on_started=start_runtime_services,
+                stream_handler=conversation_routes.stream_events,
+            )
+        finally:
+            conversation_reaper.stop_service()
 
     print(f"super-coder review layer → http://127.0.0.1:{port}  (bind {bind}, DB: {DB_PATH.name})")
     try:
--- a/.super-coder/api/server.py
+++ b/.super-coder/api/server.py
@@ -63,6 +63,7 @@ import artifact_policy  # noqa: E402
 import backfill_shell_api_keys  # noqa: E402  (startup key provisioning)
 import conversation_broker  # noqa: E402  (Feature #24 durable turn service)
 import conversation_launch  # noqa: E402  (canonical shell launch preparation)
+import conversation_reaper  # noqa: E402  (Feature #31 orphan process ladder)
 import db_driver  # noqa: E402
 import git_hygiene  # noqa: E402  (live repo dirty/stale/clean snapshot)
 import mem_credentials  # noqa: E402  (runtime Admin credential provisioning, spec #30 req 11)
@@ -4263,10 +4264,11 @@ def require_loopback_bind(bind: str) -> None:
 
 def start_runtime_services() -> None:
     """Start commit-woken conversations and the armed Sprint services."""
-    conversation_broker.start_service(
+    broker = conversation_broker.start_service(
         DB_PATH,
         launch_preparer=conversation_launch.ConversationLaunchPreparer(DB_PATH),
     )
+    conversation_reaper.start_service(DB_PATH, native_interrupt=broker.interrupt)
     sprint_runtime.start_service(DB_PATH)
     sprint_pr_watcher.start_service(DB_PATH, repo_root=REPO_ROOT)
 
--- a/.super-coder/migrations/0163_conversation_run_process_identity.sql
+++ b/.super-coder/migrations/0163_conversation_run_process_identity.sql
@@ -1,0 +1,80 @@
+-- 0163 — durable conversation-run process identity and reaper ladder.
+--
+-- Process identity belongs on the run row as recovery evidence.  The active
+-- chat registry remains the protection authority; these columns let the
+-- reaper identify and escalate an unlinked turn after its broker disappears.
+
+BEGIN;
+
+ALTER TABLE conversation_runs
+ADD COLUMN process_pid INTEGER
+    CHECK (process_pid IS NULL OR process_pid > 0);
+
+ALTER TABLE conversation_runs
+ADD COLUMN process_start_ticks INTEGER
+    CHECK (process_start_ticks IS NULL OR process_start_ticks >= 0);
+
+ALTER TABLE conversation_runs
+ADD COLUMN process_group_id INTEGER
+    CHECK (process_group_id IS NULL OR process_group_id > 0);
+
+ALTER TABLE conversation_runs
+ADD COLUMN reaper_last_signal TEXT
+    CHECK (
+      reaper_last_signal IS NULL
+      OR reaper_last_signal IN ('interrupt','SIGTERM','SIGKILL')
+    );
+
+ALTER TABLE conversation_runs
+ADD COLUMN reaper_signaled_at TEXT;
+
+CREATE TRIGGER trg_conversation_runs_process_identity_insert
+BEFORE INSERT ON conversation_runs
+WHEN NOT (
+  (NEW.process_pid IS NULL
+   AND NEW.process_start_ticks IS NULL
+   AND NEW.process_group_id IS NULL)
+  OR
+  (NEW.process_pid IS NOT NULL
+   AND NEW.process_start_ticks IS NOT NULL
+   AND NEW.process_group_id IS NOT NULL)
+)
+BEGIN
+  SELECT RAISE(ABORT, 'conversation run process identity must be complete');
+END;
+
+CREATE TRIGGER trg_conversation_runs_process_identity_update
+BEFORE UPDATE OF process_pid,process_start_ticks,process_group_id
+ON conversation_runs
+WHEN NOT (
+  (NEW.process_pid IS NULL
+   AND NEW.process_start_ticks IS NULL
+   AND NEW.process_group_id IS NULL)
+  OR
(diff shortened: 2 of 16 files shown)


jedbjorn (commented):
## S87 U3 review — PR #963 @ 332221b vs spec #89 (The Reaper / Zombie coverage / registry protection)

**Verdict: 0 Major / 1 Medium / 5 Low.** Core contract verified adversarially: protection is exact-identity-only, recycled pids are never signalled, the ladder persists across beats, the pgroup guarantees now come from the shared runner. The one Medium is a sweep-enumeration gap that silently voids the spec's zombie-class-2 promise on one reachable path.

### Attention-item verdicts (verified, not trusted)

1. **Terminal status — CONFORMANT.** The run state machine (migration 0132, trigger-enforced, mirrored in `conversation_state.py`) has no `interrupted` state; the engine-wide convention already maps semantic `run.interrupted` → storage `cancelled` (`base.py:654`). The reaper follows it exactly: `state='cancelled'` + `error_code='CONVERSATION_RUN_REAPED'` + `run.interrupted` event with a self-describing `reaper` payload block. The spec's operative clause — "never `unknown`" — is honored (see Medium 1 for the one path where it isn't). Deviation ruling: not a deviation; forking a trigger-locked vocabulary mid-unit would be the defect. See Low 1.
2. **Shutdown lifecycle — VERIFIED.** `stop_service()` stop→join→clear under the service lock; idempotent on double-stop (`_SERVICE is None` → return); safe if never started; the service spawns no other thread or timer. The heartbeat loop wakes immediately on `stop()` (`Event.wait`). Caveat: unbounded `join()` — see Low 2. The per-spawn `reap_owned_group` threads are runner-owned daemons that self-terminate; they are not reaper lifecycle.
3. **Protection rule — VERIFIED.** `_protected_sql()` matches `active.process_pid = r.process_pid AND active.process_start_ticks = r.process_start_ticks`; SQL NULL semantics make a NULL-identity registry row (opencode shared server, #961 M1) protect nothing — and the test flips a registry row to NULL and watches the run get reaped. NULL-pid run rows can never be kill candidates (`process_pid IS NOT NULL` in `candidates()`), and the new adopt filter's first disjunct (`r.process_pid IS NULL OR EXISTS(...)`) keeps them adoptable, so opencode turns partition cleanly: adoption, never the ladder. Protection is also re-checked inside every write (`eligible`, `record_signal`, `finish_interrupted` WHERE clauses), so a mid-sweep re-link can't be raced.
4. **Recycled-pid guard — VERIFIED.** Every beat re-reads `/proc/pid/stat` and matches pid+start_ticks+pgid before any ladder branch; `_signal_step` re-reads again immediately before `killpg`, including mid-ladder TERM→KILL escalation; identity mismatch finishes the run without signalling (tested). Residual kernel-level TOCTOU between the read and the `killpg` is inherent without pidfd — Low 3.
5. **Ladder / grace / pgroup / tunables — VERIFIED.** Ladder state persists on the run row (`reaper_last_signal`/`reaper_signaled_at`, pairing trigger-enforced) and escalation spans beats without blocking — tested across four beats at exact grace boundaries (14s no-op / 15s fire). 30s young grace tested at the 29s boundary with zero side effects. Decision #40 is genuinely generalized: `SubprocessRunner` now defaults `start_new_session=True` and stamps the child pid as owned pgroup (claude asserted in tests), codex's `JsonLineRpcProcess` is sessioned, kimi's private ladder is deleted in favor of the shared one — recorded pgid is always the turn's own group, so `killpg` can never hit the engine's group. Tunables are env-config with spec defaults (60/15/15/30), tested with overrides.

### Medium

**M1 — Runs finalized `unknown` with a live process are invisible to the reaper (sweep-enumeration gap).** `candidates()` enumerates only `state IN ('starting','running')`. But a protected broker-death orphan is *adopted* (the new adopt filter correctly leaves only unprotected runs to the reaper), and if reconciliation cannot prove an outcome — adapter exceptions × `reconcile_attempts`, unreadable/corrupt session state while the harness still runs — `_finish_error`/`_reconcile_loop` finalize the run `unknown` and clear the registry linkage. The OS process can still be alive; its full identity (pid+ticks+pgroup) persists on the run row; nothing will ever kill it. That is exactly spec zombie class 2 ("run rows still carry process identity… the sweep takes them") failing silently, and it also produces the `unknown`+live-process outcome the spec forbids. Suggested fix: include unprotected `state='unknown'` rows with non-NULL identity in the candidate sweep (ladder + `run.interrupted` event, no state transition — `unknown` is terminal), or route unproven-outcome adoptions to the reaper instead of finalizing. Test to add: adopted run finalized `unknown` with live identity → reaper kills it.

### Low

1. **Document the interrupted→cancelled mapping.** One spec-erratum/module-doc line stating storage `cancelled` == spec "interrupted" (and that `interrupt_evidence:"operator"` is the closest existing vocabulary for an engine-automated kill — the `reaper` payload block carries provenance) prevents every future reviewer from re-litigating item 1.
2. **`stop_service()` joins without timeout.** A beat wedged in DB-busy retry stalls server teardown indefinitely; `join(timeout)` + a warning keeps shutdown bounded.
3. **Unverifiable-kill windows, named:** (a) if the broker (and its runner reap-thread) is dead and the group leader later exits leaving children, the reaper finishes the run without signalling the group — the identity anchor is gone, so this is the safe choice, but the leak class should be documented; per-member `/proc` pgid-membership verification or pidfd_send_signal are the hardening paths for both this and the TOCTOU in item 4.
4. **`finish_interrupted` only handles a `running` conversation.** A run reaped under a `waiting` conversation (input requested) leaves the chat `waiting` with no live run until the next enqueue re-queues it; legal but unhandled — worth a comment or a `waiting` branch/test.
5. **Test gaps:** NULL-pid exclusion from `candidates()` untested; `stop_service` double-stop/never-started untested; the pending-outbox → `queued` (vs `idle`) branch of `finish_interrupted` untested; the asymmetry "broker finalize clears registry identity / reaper finalize retains run-row identity" is deliberate-looking but unasserted.

### Ruled scope honored
#959/#960/#962 not raised. Checks GREEN at 332221b confirmed; suite not re-run.

Per sprint protocol: Medium 1 gates merge — PLN1 disposition under decision #77. Comment-only review; no merge, no push.

— REV3 (Rev-03), S87 U3 review lane

jedbjorn (commented):
## S87 U3 re-review — delta @ d17a171 (base: my review @ 332221b)

**Verdict: review-clean — 0 Major / 0 Medium.** The sweep-enumeration Medium (M1, flag #161) is resolved.

### Verified
- **M1 resolved:** `_sweepable_sql()` now admits `state='unknown'` rows lacking a `run.interrupted` event, applied consistently across `candidates` / `eligible` / `record_signal` / `finish_interrupted`. The adopted-unprovable orphan class traverses the existing ladder (interrupt → SIGTERM → SIGKILL) and exits the sweep on the durable `run.interrupted` event.
- **No over-sweep:** registry-protected rows stay exempt (`_protected_sql` unchanged, applied at every gate); NULL-identity rows are filtered by `process_pid IS NOT NULL` (identity completeness enforced by the 0163 triggers); recycled-pid verify (`_matches` snapshot check) still gates every signal on the new path; proven-terminal states and unknown rows with prior `run.interrupted` evidence are excluded.
- **Diagnosis preserved:** unknown-class finalize skips run/message/conversation transitions — state stays `unknown`, `error_code=HARNESS_OUTCOME_UNKNOWN`, message `failed`, conversation `error`, process identity retained. Test asserts all of it.
- **Exactly-once, transactionally:** `finish_interrupted` re-checks `_sweepable_sql` (including the NOT EXISTS) inside the write transaction, so a double `run.interrupted` append is structurally excluded; ladder resume-by-`reaper_last_signal` is covered by the new test (sweeps 1,1,1 then 0).
- **No regressions:** delta touches only `conversation_reaper.py` + its tests. Full suite at d17a171: 1673 passed; the 3 failures (`test_migration_management`, `test_worktree_targets` ANSI-colored argparse help; `test_supervision_sc` systemd restart) reproduce identically at 332221b — pre-existing environmental, not PR regressions. Prior Lows L1–L5 untouched.

### New Lows (report-only)
- **L6 (efficiency):** the `NOT EXISTS` correlates on `conversation_events.run_id`, which has no index (only `(conversation_id, sequence)`). Finalized unknown rows keep `state='unknown'` + non-NULL pid forever, so every heartbeat pays a full event-table scan per such row, for the lifetime of the install, growing with event history. An index on `conversation_events(run_id, event_type)` would cap it.
- **L7 (projection invariant):** `run.interrupted` no longer implies message → `cancelled`. `ui/app.js:4271` live-applies that event as `message.state='cancelled'` while the DB message for an unknown-class finalize stays `failed` — transient client divergence until reload, only for a conversation already in `error`. Sprint-liveness counting the event as native evidence is pre-existing behavior extended by seconds, not a new class.

Not merged, not pushed — merge is the FnB's gate.

Issue #967, opened 2026-08-02
restart: restricted supervision test omits pre-restart database backup


jedbjorn:
Where: subfloor, DEV5 host seat, branch feat/s87-u3-reaper (the PR diff does not touch supervision or backup code).

Ran: `.venv/bin/pytest -q tests/test_supervision_sc.py::RestrictedLaunchTests::test_restart_bounces_systemd_broker_and_reports_full_inventory`

Expected: restart exits 0 and creates exactly one `shell_db.prerestart.*.db` fixture backup.

Actual: restart exits 0, but the backup glob is empty (`AssertionError: 0 != 1` at test_supervision_sc.py:394). Reproduced both in the full suite and in isolation.

Workaround: none for this unrelated test; the scoped reaper/broker/schema suites pass.

jedbjorn:
Baseline confirmation: the same isolated test fails identically at the pre-fix PR head 332221bb01d46e2335cf000d29e366dedacfc4a0, confirming the reaper review-fix diff did not introduce it.

jedbjorn:
Additional baseline repro from DEV3 S87 U8: full `./sc test` at branch `feat/s87-u8-authority-split` produced 1 failure / 1,714 passes / 1 skip, the same `test_restart_bounces_systemd_broker_and_reports_full_inventory` assertion (`0 != 1`). The isolated test also fails identically on untouched assigned base `2be3606554c3b456f32b33f989a41c354f042fa3`, whose checkout is clean. U8 changes only Sprint role skill assets, reseed migration/manifest, and skill tests. No workaround; scoped gates pass.

Issue #966, opened 2026-08-02
lint: EXE001 rejects repository Python files with non-executable shebangs


jedbjorn:
Where: subfloor, DEV5 host seat, review-fix branch feat/s87-u3-reaper.

Ran: `./sc lint .super-coder/scripts/conversation_reaper.py tests/test_conversation_reaper.py`

Expected: lint evaluates Python source consistently with the repository baseline.

Actual: Ruff exits 1 with EXE001 for both files because they have shebangs and mode 100644. Existing neighboring files such as conversation_broker.py and test_conversation_broker.py have the same shebang/mode combination, so this is broader than the current change.

Workaround: run Ruff with EXE001 ignored to assess the scoped source changes. The focused behavioral suites remain green.

jedbjorn:
Additional repro from DEV5 on branch `feat/s116-u3-pickup-exhaustion` (host seat):

- Ran: `./sc lint .super-coder/scripts/sprint_domain.py .super-coder/scripts/sprint_board.py tests/test_sprint_recovery.py tests/test_sprint_work_dispatch.py`
- Expected: targeted changed-file lint passes.
- Actual: `EXE001 Shebang is present but file is not executable` at `.super-coder/scripts/sprint_domain.py:1`. The branch does not change file mode.
- Workaround: treated as pre-existing baseline; relied on full `./sc test` for the scoped PR.

PR #974, opened 2026-08-03
feat(sprints): add coordinate routing and FnB close

jedbjorn:
## Summary

- persist Sprint-scoped coordinate mode and project its audit events
- treat planner-bound Re-enter messages as New while coordinate mode is active, preserving busy-turn delivery at the natural boundary
- make the FnB close path immediate and universal, including Sprint chats and active runs, with registry unlinking for reaper ownership
- clear coordinate mode only for FnB pause, cancel, or Sprint close actions
- pin all declared Sprint routing literals with regression coverage

## Verification

- focused routing/conversation/board suites: 137 passed plus 40 subtests
- migration/schema/snapshot gates: 64 passed plus 295 subtests
- full suite: 1692 passed, 1 skipped, 1 pre-existing failure
- new coordinate suite: 6 passed; Ruff clean

## Known baseline

- #967: test_supervision_sc restart fixture does not create the expected shell_db.prerestart backup
- #966: repository-wide Ruff format/lint baseline remains red; no unrelated formatting churn included

Sprint 87 U5 · task #264. Never merge without authorization.

### DIFF
--- a/.super-coder/api/conversation_routes.py
+++ b/.super-coder/api/conversation_routes.py
@@ -586,7 +586,46 @@ def _cancel_queued_turns(con, conversation_id: str) -> int:
     return int(cancelled)
 
 
-def _begin_close(con, conversation_id: str, current_state: str) -> int | None:
+def _enable_planner_coordinate_mode(
+    con,
+    *,
+    shell_id: int,
+    conversation_id: str,
+) -> int | None:
+    sprint = con.execute(
+        "SELECT sprint_id FROM sprints WHERE lifecycle='armed' "
+        "AND originating_planner_shell_id=? AND coordinate_mode=0",
+        (shell_id,),
+    ).fetchone()
+    if sprint is None:
+        return None
+    sprint_id = int(sprint["sprint_id"])
+    changed = con.execute(
+        "UPDATE sprints SET coordinate_mode=1,updated_at=datetime('now'),"
+        "version=version+1 WHERE sprint_id=? AND lifecycle='armed' "
+        "AND coordinate_mode=0",
+        (sprint_id,),
+    ).rowcount
+    if changed != 1:
+        return None
+    con.execute(
+        "INSERT INTO sprint_events "
+        "(sprint_id,event_type,actor_kind,payload) VALUES "
+        "(?,'coordinate_mode.enabled','fnb',?)",
+        (
+            sprint_id,
+            json.dumps(
+                {"conversation_id": conversation_id, "reason": "Planner chat closed"},
+                separators=(",", ":"),
+                sort_keys=True,
+            ),
+        ),
+    )
+    return sprint_id
+
+
+def _begin_close(con, conversation_id: str, current_state: str) -> None:
+    """Unconditionally close and unlink a chat selected by the FnB."""
     active = con.execute(
         "SELECT run_id,trigger_message_id FROM conversation_runs "
         "WHERE conversation_id=? AND state IN ('leased','starting','running') "
@@ -604,26 +643,6 @@ def _begin_close(con, conversation_id: str, current_state: str) -> int | None:
                 {"cancelled_queued_turns": cancelled},
                 run_id=run_id,
             )
-        interrupt_exists = con.execute(
-            "SELECT 1 FROM conversation_events WHERE run_id=? "
-            "AND event_type='run.interrupt.requested' LIMIT 1",
-            (run_id,),
-        ).fetchone()
-        if interrupt_exists is None:
-            _append_event(
-                con,
-                conversation_id,
-                "run.interrupt.requested",
-                {"reason": "conversation.close"},
-                message_id=int(active["trigger_message_id"]),
-                run_id=run_id,
-            )
-        con.execute(
-            "UPDATE conversations SET last_activity_at=datetime('now'),"
-            "version=version+1 WHERE conversation_id=?",
-            (conversation_id,),
-        )
-        return run_id
 
     recovered_state = None
     if current_state == "queued":
@@ -632,7 +651,8 @@ def _begin_close(con, conversation_id: str, current_state: str) -> int | None:
             (conversation_id,),
         )
     elif current_state == "running":
-        recovered_state = current_state
+        if active is None:
+            recovered_state = current_state
         con.execute(
             "UPDATE conversations SET state='error' WHERE conversation_id=?",
             (conversation_id,),
@@ -641,6 +661,10 @@ def _begin_close(con, conversation_id: str, current_state: str) -> int | None:
         "UPDATE conversations SET state='closed',closed_at=datetime('now'),"
         "last_activity_at=datetime('now'),version=version+1 "
         "WHERE conversation_id=?",
+        (conversation_id,),
+    )
+    con.execute(
+        "DELETE FROM active_shell_chats WHERE chat_id=?",
         (conversation_id,),
     )
     _append_event(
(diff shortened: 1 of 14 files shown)


jedbjorn (commented):
## S87 U5 review — PR #974 @ 6e50639 vs spec doc #89 (routing table / coordinate flag / FnB close)

**Verdict: 0 Major / 1 Medium / 2 Low.** Medium gates merge per sprint contract. Comment-only review — merge is PLN1's under decision #77.

### Verified (attention items)
1. **Routing literals** are declared at send sites only: work_assignment `new` (`sprint_domain.py:2388`), review_request `new` (`sprint_review_loop.py:96`), review changes `re-enter` (`sprint_review_loop.py:161`), decline→planner `re-enter` (`sprint_message_delivery.py:469`), merge-bypass→planner `re-enter` (`sprint_domain.py:2308`), relay `re-enter`. The coordinate CASE in `claim_next` is a declared-type transform *upstream* of delivery resolution, not a re-implementation of it — the live-process check in `_resolve_conversation` still governs mid-turn.
2. **Coordinate flag is tracked** (`sprints.coordinate_mode`, migration 0165), set only on FnB close of the armed sprint's planner chat, cleared only by an `fnb`-actor pause/cancel/close (`_clear_coordinate_mode` guards `actor.kind`). Auto-pause preserves it; a planner-actor abort preserves it; FnB confirm-pause on an already-paused sprint clears it (the spec's pause+resume way back). All four pinned by tests. Global single-armed-sprint index (`idx_sprints_single_armed`) makes the `fetchone()` in `_enable_planner_coordinate_mode` safe.
3. **Displacement race (flag #142 heritage):** `_resolve_conversation` short-circuits on `has_live_process` before `wants_new`, so a coordinate-flipped New re-enters the active chat when the planner is mid-turn. `test_coordinate_mode_new_reenters_when_planner_is_mid_turn` exercises a real `/proc`-identity-verified live pid and asserts no rotation. Boundary timing itself is the 0164 queued-turn machinery, unchanged here.
4. **FnB close unconditional:** the `SPRINT_CONVERSATION_MANAGED` PATCH guard is gone, `_begin_close` closes + cancels queued turns + unlinks the registry in one transaction, and the displaced run is left to the reaper — which is exactly what the spec prescribes. Broker `finish_run` now tolerates an already-closed conversation without resurrecting its state; a dispatched run's trigger message is not touched by `_cancel_queued_turns` (outbox already `dispatched`), so no illegal message transitions. The POST-message reopen guard for sprint chats survives (`test_closed_sprint_conversation_never_reopens`).
5. **No unmerged-lane assumptions:** the diff leans only on 0162 (registry + close trigger), 0163, 0164, and its own 0165; `merge_grant_enabled`, `sprint_events.actor_kind`, `originating_planner_shell_id` are all on main. 0165 is additive with default 0 and matches the repo's BEGIN/COMMIT migration convention; the removal-manifest update is mechanical.

### M1 (Medium — spec conformance): shell-scoped planner notices bypass coordinate mode
`_queue_planner_notice` (`sprint_domain.py:1616`) sends via `send_to_shell_in_transaction`, which stores `sprint_id=NULL, to_participant_id=NULL` (`sprint_message_delivery.py:196-211`). The coordinate CASE (`sprint_message_delivery.py:744-746`) requires `p.role='planner' AND s.coordinate_mode=1`, so these planner-bound `re-enter` messages can **never** flip to New. This is reachable while armed with the flag set: the resume reconciliation notice (`sprint_domain.py:460`) fires immediately after auto-pause→resume — precisely the flow whose flag preservation the spec mandates — and delivers as an unflipped Re-enter into the planner's stale idle ticket chat instead of opening a fresh one. Spec callout: *"while set, Re-enter messages to the planner behave as New."* The hole is structural: any future engine-authored planner send through `send_to_shell` inherits it. Proposed fix: key the flip on `receiver_shell_id` = the armed sprint's planner shell (so scope-less messages participate), or participant-scope the resume notice.

### L1 (Low — dead code): the close-interrupt path is unreachable
`conversation_routes.py:1118` sets `active_run_id = None` and nothing ever reassigns it, so `:1202-1203` never fire and `_deliver_close_interrupt` is dead. Reaper-only kill is spec-conformant, so behavior is fine — but the vestige tells readers close still interrupts the run. Delete it (or wire the nudge deliberately).

### L2 (Low — implicit invariant): coordinate enable is keyed on the shell, not the closed chat's registry row
`_enable_planner_coordinate_mode` flips on FnB close of *any* chat owned by the planner shell. That is safe today only because 0162's backfill + close trigger guarantee at most one open (and registered) chat per shell. If any future path leaves an open unregistered chat, closing it would silently flip the mode. Cheap hardening: `_begin_close` already executes the registry DELETE — check its rowcount (or compare against `active_shell_chats`) before enabling.

### Nits (no gate)
- Duplicate adjacent `if target == "completed":` blocks in `sprint_domain.transition`.
- The CASE's `s.lifecycle='armed'` predicate is redundant with the claim gate at `sprint_message_delivery.py:719` (harmless).
- The wake prompt renders "(declared New)" for a coordinate-flipped message — it is the effective type, not the sender's declaration; consider "(delivering as New)".
- Explicit `DELETE FROM active_shell_chats` in `_begin_close` duplicates 0162's close trigger (harmless belt-and-braces).
- No test that closing a dev/rev chat mid-armed-sprint leaves `coordinate_mode` untouched.

jedbjorn (commented):
## Re-review @ 2e194e6 (S87 U5 re-review, delta from 6e50639) — REV1

**Verdict: clean — 0 Major / 0 Medium.**

**M1 resolved** (8e07fe4, `sprint_message_delivery.py` claim_next): the coordinate CASE is now keyed on the receiver — `EXISTS (… armed coordinate sprint WHERE originating_planner_shell_id = m.receiver_shell_id)` — independent of message sprint scope. The M1 gap (NULL-scope engine notices, e.g. the resume-reconciliation notice reachable while armed + flag set, bypassing the flip) is closed: `test_coordinate_mode_rotates_engine_scoped_planner_notice` proves the old planner chat closes, a fresh chat opens with `(declared New)`, and the stored row keeps `sprint_id=NULL` / `re-enter` — U2 F4 (engine notices stay unscoped) preserved. No over-catch: both sides of the EXISTS predicate are NOT NULL (`wake_message.receiver_shell_id` per 0164, `sprints.originating_planner_shell_id` per 0146), and only the originating planner shell of an armed coordinate sprint can match — a non-coordinate planner or non-planner receiver falls through to the literal `declared_type`. Flag-clear pinned by `test_engine_scoped_planner_notice_reenters_when_mode_is_clear`.

**Rebase/renumber sound**: 0166 body is byte-identical to the reviewed 0165 except the header line; `tests/fixtures/sprint_removal/manifest.json` correctly lists U6's `0165_pr_subscriptions.sql` before `0166_sprint_coordinate_mode.sql`; no duplicate numbering introduced by this PR (the pre-existing 0155 pair is already on main — out of scope here, flagged upstream separately). The rebase reconciled cleanly against U6's concurrent edit to the same file (`SprintMessageStore.resolve_in_transaction`) — orthogonal to claim_next; a full diff-of-diffs vs the reviewed round shows no other content drift.

**L1 / L2**: `conversation_routes.py` is content-identical to the reviewed round (same blob `274d1f8..870cf88`) — L1 (dead close-interrupt path) stays report-only per dev disposition; tolerable. L2 unchanged.

**Tests/CI**: all PR-touched test files pass locally at head (155 passed + 27 subtests); CI fully green at 2e194e6. One local-only failure in `test_migration_management.py` (help-text assertion vs colorized argparse) reproduces identically on main and passes in CI — host environment, not this PR.

**Non-blocking note (Low)**: scope-independence means a shell originating an armed coordinate sprint has *all* its Re-enters flipped, including any scoped to a different sprint it might participate in under a future dual-role arrangement. Structurally safe under today's fixed one-role shells; no regression pins "non-planner receiver unaffected while a coordinate sprint is armed" — worth a test if dual-role ever lands.

Merge remains PLN1's gate.

Issue #973, opened 2026-08-03
restart: restricted systemd test does not create prerestart backup


jedbjorn:
Observed on host seat in DEV5 while verifying Sprint 87 U4.\n\nCommand: `.venv/bin/pytest -q tests/test_supervision_sc.py::RestrictedLaunchTests::test_restart_bounces_systemd_broker_and_reports_full_inventory`\n\nExpected: the fixture's restart command creates exactly one `shell_db.prerestart.*.db` backup.\n\nActual: restart exits 0, but the backup glob is empty (expected 1, got 0). The exact test fails both on untouched base `2537e49` and on the U4 feature worktree; the other 1693 full-suite tests pass on U4.\n\nWorkaround: none applied; treated as a verified pre-existing baseline failure and excluded from the U4 regression judgment.

Issue #263, opened 2026-07-03
[engine] vm-broker: no seam to reach the guest's Windows-MCP endpoint from a sandboxed seat (GUI-driving follow-up to #262)


jedbjorn:
Follow-up to #262. The `windows_vm_gui` skill (GUI-driving the Windows Test VM via Windows-MCP / UIA) lands with its per-session connect written for **host-run seats only** — `ssh -f -N -L 18000:127.0.0.1:8000` + `claude mcp add --transport http`. That path cannot work from the sandbox, by design:

- The sandbox has no `ssh`, no key, and no route across libvirt NAT (see `.super-coder/docs/windows-vm-broker.md` — "The gap (measured)").
- A host-side tunnel bound to `127.0.0.1` is the host's loopback, not the container's.
- Binding a tunnel on the docker-bridge gateway was **explicitly rejected** in the broker spec ("a real network surface that then needs its own auth token. The socket avoids that entirely").

So sandboxed dev/reviewer seats — the engine's default — can prep the guest (broker `/exec`) and take screenshots (`/capture`), but cannot hold a live MCP session against the guest's Windows-MCP server. The skill documents the gap and points here.

## What's needed

A broker-mediated path from the sandbox to the guest's `streamable-http` MCP endpoint, consistent with the unix-socket posture. Candidate designs, in rough order of fit:

1. **Streaming reverse-proxy route on the vm-broker socket** — broker opens/owns an `ssh -N -L` to the guest's localhost-bound Windows-MCP and proxies `/mcp/*` over the existing unix socket. Needs: chunked/SSE streaming through the `BaseHTTPRequestHandler` broker (non-trivial), plus a tiny in-sandbox TCP→unix relay because `claude mcp add --transport http` only speaks TCP URLs.
2. **Tunnel sidecar container on `sc-net`** — a minimal host-launched container running only the ssh forward, reachable by container DNS like the pg sidecar. Needs: the key mounted into the sidecar (posture change — today the key never leaves the host process), plus an auth story (`windows-mcp serve --auth-key`) since `sc-net` is shared across forks.
3. **Gateway-bind tunnel + auth key** — smallest code, but re-opens the network-surface decision the broker spec closed; would need the auth key distributed via the fork-gated socket (`GET /vm`).

[...]

PR #975, opened 2026-08-03
feat(wakes): add activity ceiling and retry recovery

jedbjorn:
## Summary\n- apply 15/60/180-second `available_at` backoff to the three-attempt wake budget\n- close and unlink the target active chat on terminal failure before the existing `wake_delivery_exhausted` auto-pause path\n- add an engine-wide mechanical activity monitor with `SC_CHAT_INACTIVITY_CEILING` (seconds, default 3600) and dispatch queued wakes after hung chats close\n- requeue terminal-failed NULL-sprint wakes on the monitor pulse, closing review R6\n\nNo migration is required; the existing wake `available_at`, registry, run-evidence, and join schema are sufficient. System auto-pause remains distinct from FnB lifecycle actions, leaving U5's coordinate-flag clear policy untouched.\n\n## Verification\n- focused: 90 passed, 3 subtests passed\n- targeted ruff + diff check: clean\n- full `./sc test`: 1693 passed, 1 skipped, 1 pre-existing failure\n- base proof: the same supervision restart-backup test fails on untouched `2537e49`; tracked in #973


jedbjorn (commented):
## S87 U4 review — task #263, spec doc #89 (Activity Monitor / Reaper seams / Zombie coverage) @ cbb2fa4

**Verdict: 0 Major / 1 Medium / 4 Low.** The Medium gates merge (flag SC-010). The five attention items from the assignment were each walked against the code; results below, findings after.

### Attention-item verification

**1. Budget rewire chain — VERIFIED, transactional-sane.** The entire terminal hop runs in one write transaction (`sprint.wake_failure`): attempt evidence insert → outbox `failed` → `close_for_inactivity` (conversation flipped, queued/claimed turn messages cancelled, `close_active` sets `closed` + deletes the registry row with a fail-loud unlink check) → `conversation.closed` event → message-scope-keyed sprint pause. `conversation_events.notify` fires only after commit. There is no closed-chat-but-linked-registry window for the reaper to misread. U2's scope-keyed pause accounting (armed sprints joined via `wake_message.sprint_id`, not wake scope) is untouched context — no regression. The `target_conversation_id` guard prevents closing a replacement active chat (covered by `test_terminal_failure_does_not_close_a_replacement_active_chat`).

**2. Backoff — VERIFIED.** `WAKE_DELIVERY_BACKOFF_SECONDS = (15, 60, 180)`, indexed by attempt with the pre-existing `attempt > 3` invariant guard ahead of it (no index overrun possible). Budget is exactly 3 (`terminal = attempt == 3`). The claim gate (`state='pending' AND available_at<=now`) already existed; the defect was that failures never advanced `available_at` — this fix is minimal and correct. Coalescing untouched (`sprint_message_delivery.py` has no non-test changes). Tests measure the actual 15/60/180 deltas between `attempted_at` and `available_at`.

**3. Ceiling vs long turn — the walk.** `last_activity_at` **does** advance during a legitimately long-running turn: `BrokerStore.append_events` updates it on every stream flush, and the broker flushes immediately on any meaningful event (tool.started, tool.completed, message boundaries) and on a short deadline for assistant deltas. A long test suite run as tool calls therefore keeps the chat fresh — the named scenario is safe. The residual hole is a **single silent tool execution**: the claude adapter emits nothing between `tool.started` and `tool.completed`, and `renew_runs` refreshes only `conversation_runs.heartbeat_at`, never `last_activity_at`. One tool call silent for > 3600s (unbounded MCP call; a non-claude adapter with a >1h exec timeout — claude's Bash cap is 10min so it can't hit this) gets a healthy turn closed and its process reaped. Filed as **L1**, severity per plausibility (low on this host's harness set). Everything else in item 3 verified: engine-wide on the 5s pulse ✓, default 3600 / `SC_CHAT_INACTIVITY_CEILING` validated ✓, keyed to `conversations.last_activity_at` ✓, queued wake dispatches in the same pulse as the close (`test_runtime_closes_hung_chat_then_dispatches_queued_wake`) ✓, protected-row paradox resolved — close clears the registry row in the same transaction and the reaper sweep test then takes the pid ✓.

**4. R6 fold-in — per-wake once-only VERIFIED; chain is NOT once-only → M1 (Medium).** A recovered wake is never re-selected (the join-move empties its `EXISTS`), and the second tick recovers nothing (tested). Drain composition is safe: delivery drains on `delivered_at IS NULL` per receiver, so a message rescued by any other wake turn is never double-delivered. But see M1 below.

**5. Coordinate-flag preserve — VERIFIED.** Zero references to coordinate state anywhere in the diff; the terminal path only adds the chat close ahead of the untouched pause call. U4 assumes nothing from U5's lane.

---

### M1 (Medium) — engine-wake recovery loops forever when the recovery wake itself terminal-fails

`_recover_engine_wakes` selects `sprint_id IS NULL AND state='failed'` wakes with undelivered NULL-sprint messages, but nothing excludes wakes it created itself. A recovery wake carries `sprint_id NULL`; if the receiver still cannot boot, it burns its own 3-attempt budget (15/60/180), goes `failed`, and the next pulse recovers it again under a fresh key (`engine-recovery:failed-wake:{new_id}` — the key changes every cycle, so idempotency never terminates the chain). Result: an infinite ~8-minute cycle per stuck shell — each cycle inserts a wake row, 3 attempt rows, a fresh wake conversation (new wake_id → new `generation:wake:` key) and then closes it via the terminal path — with **no surface at all** (NULL-sprint failures pause nothing). That is unbounded background churn (~180 conversations/day per stuck shell) and violates the ruled requirement that terminal-failed NULL-sprint wakes move **once** to a delayed recovery wake. `test_failed_engine_wake_is_recovered_once_after_terminal_backoff` proves per-wake idempotence only; the chain case is untested. Fix direction: skip wakes whose `idempotency_key LIKE 'engine-recovery:%'` (one recovery shot, then strand-with-evidence), or bound the chain and surface a durable notice to the FnB after the second terminal failure.

### L1 (Low) — a single silent tool execution longer than the ceiling kills a healthy turn

Walk in item 3 above. When it fires, the consequence is softened by the reaper ladder (harness-native interrupt → TERM 15s → KILL 15s), but the operator evidence is only a `conversation.closed` event on a chat that was doing real work. Cheap mitigation: if the newest conversation event is a `tool.started` without a matching `tool.completed`, require a second, larger ceiling before closing — or document that forks with long-running MCP/exec tools must raise `SC_CHAT_INACTIVITY_CEILING`.

### L2 (Low) — sprint-runtime thread dies half-silently on a malformed `SC_CHAT_INACTIVITY_CEILING`

In `run()`, `ActivityMonitor(...)` (and its `from_env` validation) is constructed *outside* both try blocks. A typo'd env var raises `ValueError`, the daemon thread dies, and the `finally` sets `_started_once` — so `wait_started()` returns `True` while the entire wake/sprint runtime is dead. Validate the config before thread start (fail loud at service construction) or move construction inside the guarded loop.

### L3 (Low) — `close_for_inactivity` duplicates `close_for_wake` and is a misnomer on its second call site

It is a line-for-line copy of `close_for_wake` minus the `has_live_process` guard (with a slightly drifted outbox-cancel formulation), and `record_wake_failure` calls it for the wake-exhaustion close, where "after the mechanical inactivity ceiling" (its docstring) is false. One shared displacement helper with an `allow_live` flag — or a neutral name (`close_for_displacement`) — keeps a single code path for a state transition this delicate.

### L4 (Low) — recovery has no age bound: historical failed wakes resurrect on deploy

The recovery scan has no cutoff, so the first monitor pulse after this lands will resurrect and deliver every pre-existing failed NULL-sprint wake with undelivered, unread messages — stale bodies arriving as fresh wakes. Likely a small set on this install, but worth either an age bound or a conscious one-time acceptance at deploy.

### Observations (no gate)

- A fresh message coalescing into a pending wake mid-backoff inherits up to 180s of delay — inherent to per-wake backoff, bounded, fine.
- `activity_monitor.py` rides the `sprint_removal` manifest, so sprint-removed forks lose the hung-turn ceiling while keeping registry + reaper. Defensible (the recovery half needs sprint tables) but worth FnB awareness: zombie class 3 reopens on those forks.
- Checks GREEN verified (tests, verify, render-check, CodeQL). The manifest fixture fix (eddf30a→cbb2fa4) is mechanical, as described.

Comment-only review — merge is PLN1's gate (decision #77). Flag SC-010 (#164) tracks M1.

jedbjorn (commented):
## S87 U4 re-review (REV2) — delta @ 65b9f47 — CLEAN (0 Major / 0 Medium)

Re-review of the fix commit 65b9f47 against my prior review @ cbb2fa4 (msg #1324, flag SC-010/#164). All four addressed findings verified against the actual code paths, plus a regression pass over the previously-clean items. Full suite run at head: 1698 passed; the 2 failures (`test_migration_management` help-text ANSI color, `test_supervision_sc` systemd restart) are host-environment artifacts that also fail at merge-base 2537e49 — not PR regressions.

### M1 — once-only recovery: VERIFIED FIXED
Walked the identity-exclusion keying end to end, not just the test:
- Outbox `idempotency_key` is **engine-generated at every insert site** — callers never touch it. Engine-scoped (`sprint_id IS NULL`) wakes get exactly two key shapes: `receiver:<shell>:wake-for-message:<id>` (`_coalesce_wake`) and `engine-recovery:failed-wake:<id>` (the scan). Sprint-scoped inserts (`sprint-resume:`, `sprint-recovery:`) are outside the scan's `sprint_id IS NULL` filter. No path can forge or collide with the `engine-recovery:` prefix.
- Schema backs the filter: `idempotency_key TEXT NOT NULL UNIQUE` (no NULL escape from `NOT LIKE`), `created_at NOT NULL DEFAULT`, and the `one-pending-per-receiver` partial unique index structurally prevents concurrent recovery wakes per shell (the scan's pending-reuse branch respects it).
- Terminal recovery wake: `record_wake_failure` raises `SprintInvariantError` on any already-terminal wake, so the exhausted branch fires **exactly once per wake** — flag + event cannot duplicate. Evidence retained: the failed wake row stays (excluded from the scan forever), and undelivered messages stay attached (test asserts `(replacement, None, None)` and outbox count 2).
- FnB surface actually fires: `INSERT INTO flags (display_name,priority,description,shell_id)` fits the schema (`priority='High'` passes the CHECK, `resolved` defaults 0), and it is **outside** the `if closed is not None` guard — the flag lands even when no active chat exists to close. The `conversation.closed` event with `unbootable_shell: true` fires only when a chat was closed, which is the correct scoping.
- Chain bound: messages can hop through a genuinely-new ordinary pending wake (which is then itself recoverable once), but every hop requires an externally-created wake; the recovery system can never feed itself. Recovery-of-recovery cannot fire.

### L4 — deploy-resurrection bound: VERIFIED
`created_at >= datetime('now','-86400 seconds')` on the original failed wake; `created_at` is NOT NULL. Regression test (2-day-old wake, terminal, untouched) confirms.

### L2 — ceiling fallback: VERIFIED
`_configured_ceiling` now catches ValueError and rejects nonfinite/nonpositive via `math.isfinite`, warns, and returns 3600. Covers `0`, `-1`, `later`, `nan`, `inf`. The thread-survival regression (`test_malformed_ceiling_does_not_kill_runtime_thread`) starts a real `SprintRuntimeService` under a malformed env and asserts it stays alive.

### L3 — closure unification: VERIFIED, live guard correct at all three sites
`close_for_displacement(allow_live_process=)` is a faithful union of the two prior bodies (the outbox-cancel by conversation_id vs by message_ids is set-identical inside the single write transaction). Call-site audit:
- `sprint_message_delivery.py:866` (New delivery) → `close_for_wake` → `allow_live_process=False`, still raises `ActiveChatBusy` on a live turn, **checked before** any queued-work cancellation — the guard is not weakened.
- `activity_monitor.py:130` (inactivity ceiling) → `close_for_inactivity` → allow-live, as designed (displace the live-but-silent turn; reaper takes the pgid).
- `sprint_domain.py:723` (terminal wake failure) → explicit `allow_live_process=True`, matching prior behavior, still scoped by the `target_conversation_id` match guard.
No other callers exist.

### Regression pass on previously-clean items
Terminal-chain transactionality (flag + event + close + pause all inside `sprint.wake_failure` write transaction), scope-keyed pause, 3-attempt budget/backoff constants, same-pulse queued dispatch, zero coordinate refs — all unchanged by the delta. Sprint-scoped wakes keep the original `"wake delivery exhausted"` event reason (`recovery_exhausted` requires `sprint_id IS NULL`).

### Residual Lows (non-blocking, report-only)
1. **Finite-overflow ceiling silently disables the monitor** (pre-existing, adjacent to L2): a huge-but-finite value (e.g. `SC_CHAT_INACTIVITY_CEILING=1e300`) passes the isfinite/positive validation, but SQLite returns NULL for `datetime('now','-1e+300 seconds')`, so `_close_inactive_chats` matches nothing, silently. Verified against SQLite directly. A sane upper bound (or a NULL-check on the rendered modifier) would close it.
2. **Age-excluded wakes have no FnB surface** (accepted L4 design, noting the asymmetry): a >24h failed engine wake's undelivered messages stay stranded with no flag — only budget-exhausted recovery wakes get one. Fine as the deploy-resurrection tradeoff; worth remembering operationally.
3. L1 (silent >1h in-tool execution) remains report-only per prior agreement.

**Recommendation: approve for merge** (FnB gate). Flag SC-010/#164 resolved on the strength of this verification.

PR #976, opened 2026-08-03
feat(pr): engine-wide shell-owned subscriptions

jedbjorn:
## Summary
- add first-class PR subscriptions keyed to the owning Developer shell, with authenticated `sc pr subscribe` support outside Sprints
- make the installation watcher poll whenever subscriptions exist, independent of Sprint lifecycle, while preserving six-state normalization and hash-chained dedupe
- route red/green/closed events only to the owning Developer as engine-wide Re-enter wake messages with repository, PR number, head SHA, event, and action
- keep merges programmatic, remove watcher-originated Reviewer handoff, and move stale-message wake cancellation behind the message store
- snapshot/rebuild subscription state and retain Sprint registration/transition compatibility projections

## Verification
- `.venv/bin/pytest -q --deselect tests/test_supervision_sc.py::RestrictedLaunchTests::test_restart_bounces_systemd_broker_and_reports_full_inventory`
  - 1694 passed, 1 skipped, 1 deselected, 869 subtests
- `./sc render-check`
- focused watcher/review/recovery/CLI/snapshot gate: 83 passed, 9 subtests
- unfiltered `./sc test`: only the pre-existing #967 supervision backup fixture fails

## Notes
- built from required base `2537e49`
- migration `0165_pr_subscriptions.sql`; U4/U5 are parallel migration lanes, so later merge order may require renumbering
- never merge: Sprint 87 manual Planner gate owns review and merge


jedbjorn (commented):
## S87 U6 adversarial review — PR #976 @ a51f0cc vs spec doc #89 (PR Subscriptions + Retired Machinery)

**Verdict: 1 Major / 0 Medium / 2 Low.** The Major gates merge (Medium+ gate).

---

### MAJOR M1 — No subscription lifecycle: observation is permanent, merged PRs poll forever, and there is no off-switch

`SprintPRWatcher.poll_once` (`sprint_pr_watcher.py:363-371`) selects every `pr_subscriptions` row unconditionally, and nothing in the engine ever removes, deactivates, or quiesces a subscription:

- A **merged** PR is terminal on GitHub (cannot reopen), yet its subscription keeps costing GitHub reads every 5s pulse for the life of the installation — `_observe_rows` groups into one `gh pr list --state all` per repo when >1 subscription, but that is still ≥1 gh call per subscribed repo per pulse, forever, even when the engine is fully idle. One repo ≈ 720 calls/hour of the host's shared 5000/hr gh quota; rate-limit contention then degrades *live* PR observation through the 60s rate backoff.
- The state is **irreversible through the engine**: `pr_subscription_transitions` is append-only (DELETE trigger aborts) and FK-references `pr_subscriptions` with `PRAGMA foreign_keys=ON` (`db_driver.connect`), so a subscription with any history cannot be deleted from a normal engine connection. There is no `unsubscribe` verb on the CLI or API. Even the FnB has no remedy short of trigger-disabling SQL surgery.
- `test_terminal_subscription_stays_observed_without_duplicate_wakes` codifies eternal observation as intended, so this is also a **spec gap**: doc #89 retires the armed-only gate ("subscriptions are engine-wide") but is silent on subscription end-of-life — and the armed gate *was* the old quiescence mechanism. Nothing replaced it. The old watcher went quiet when the sprint closed; the new one can never go quiet again after the first subscription.
- Related growth: with an unreachable repo (deleted/renamed), `_poll_failed` appends an undeletable `pr_subscription_poll_failures` row every ≤300s per subscription, forever.

Minimal fix preserving every spec behavior: skip subscriptions whose latest transition is `merged` in `poll_once` (terminal, zero-value polling); decide the closed-without-merge policy separately (closed PRs *can* reopen); and/or add an authenticated unsubscribe. Needs a Planner/FnB ruling on intended subscription lifecycle either way.

### LOW L1 — dead column selection left over from the retired reviewer routing

`_route_transition` still selects `u.reviewer_shell_id` (`sprint_pr_watcher.py:592`) but no longer reads it anywhere — it fed the deleted head-change delta-review path. Drop it so the query states what the routing actually uses.

### LOW L2 — `/_sc/pr/subscribe` surface auth is untested at the endpoint level

Ownership (dev-flavor) and authenticated-identity resolution are tested at store and CLI level, but the endpoint's 401 paths (missing/invalid bearer) and the non-dev error mapping through `_sprint_error` have no direct test; the R6-pattern consistency check is currently by inspection only.

---

### Verified against the six attention items (all clean)

1. **Routing foundation replaced, not wrapped.** Wake targeting is keyed solely on `pr_subscriptions.owner_shell_id`; the old role×state `recipients` table and the reviewer head-change delta-review path (participant lookup, `review.requested` event, `in_review` flip) are gone. Head-change after approval now invalidates the stale approval and flips `merge_ready→fixing` ('fixing' is valid disposition vocabulary, 0146) with **no** reviewer wake — `test_head_change_invalidates_approval_without_waking_reviewer` asserts shell 2 receives zero `pr-%` messages. No role-conditional survives in the watcher's wake targeting.
2. **Always-on gate removal.** All armed-lifecycle checks are out of observe/poll/poll-failure; the service starts unconditionally and `poll_once` performs zero GitHub calls with no subscriptions (tested). Polling (5s pulse), exponential + rate backoff, six-state `normalize_state` (docstring-only diff vs old), and the hash-chained transition dedupe carry over unchanged.
3. **Event semantics.** red/green/closed route to the owning dev only, all `declared_type='re-enter'`; every body is self-describing (`repository=…, number=…, head_sha=…, event=…`) — asserted in watcher, engine-wide, and CLI tests, so post-rotation cross-unit deliveries name their PR. Merged emits no wake and stays programmatic via `review_loop.observe_merge_in_transaction` (decision #36 preserved); the old closed-without-merge Planner notification is gone; no Planner/Rev git wake exists anywhere in the new watcher.
4. **Cancellation through the message store.** The hand-rolled read-mark + `sprint_wake_outbox` cancellation inside `_invalidate_stale_approval` is retired; the watcher now calls `SprintMessageStore.resolve_in_transaction`, which reuses the store's `_cancel_resolved_wakes` invariant. No outbox writes remain in the watcher; the approval wake is asserted `cancelled` in tests.
5. **New surface.** `_pr_post` follows the R6 pattern exactly (`_require_shell_auth` → early return, exact-path 404 allowlist, `_sprint_error` mapping, con closed in finally), routed before `/_sc/sprint/`; `sc pr` dispatcher + help entries added; `pr_cli.py` mirrors `sprint_cli.py`'s `mem._api` precedent; dev-only ownership is enforced in the store and the sprint `register` path attaches to an existing engine-wide subscription with ownership-conflict invariants.
6. **Migration 0165.** No collision with main (main tops at 0164; renumber vs U4/U5 lanes is a lane mechanic, not a finding). Backfill projects existing registrations and their transition history into the new tables; append-only triggers guard both evidence tables; snapshot `SPRINT_INSTANCE_TABLES` and the rebuild test cover all three new tables. Recovery correctly re-keys observation to `subscription_id`, and the resume merge-projection change is consistent and documented in the test.

Suite green at head (hosted checks). Review is comment-only — merge is PLN1's under decision #77.

jedbjorn (commented):
## RE-REVIEW — S87 U6 @ 6692a9b (delta from a51f0cc)

Delta is a single commit (`fix(pr): quiesce terminal subscriptions`) touching `sprint_pr_watcher.py` + its tests. Verdict: **review-clean — 0 Major / 0 Medium / 1 Low**.

### M1 — resolved per decision #80 (derived quiescence) ✅
- **Skip derives from the transition record, not a fresh API call**: `poll_once` filters subscriptions via a correlated subquery on `pr_subscription_transitions` (latest `normalized_state NOT IN ('merged','closed')`). Pure DB read; both new tests assert `get_calls` is unchanged on pulse for merged- and closed-latest subscriptions.
- **Re-subscribe reactivates**: `subscribe()` now observes on every call (dropped the `receipt.created` guard), so a repeat subscribe snapshots the PR; a reopened PR records a non-terminal transition and rejoins the pulse set. `test_closed_subscription_quiesces_until_resubscribed_after_reopen` proves the full cycle — quiesce → re-subscribe (same subscription_id, `created=False`) → transition ledger `[closed, green]` → next pulse polls and wakes.
- **No over-quiescence**: `COALESCE(...,'')` keeps never-observed subscriptions eligible, and `normalize_state`'s vocabulary confirms `merged`/`closed` are exactly the terminal states — `red`/`green`/`pending`/`created` still poll. The service loop pulses on a timer and ignores `poll_once`'s return, so an all-terminal table cannot wedge the watcher; reactivation is picked up on the next pulse.
- **Ruling constraints held**: no migration, no unsubscribe verb, transition insert path (hash chain) untouched by the delta.

### L1 — resolved ✅
`reviewer_shell_id` no longer appears anywhere in the watcher; the `unit_rows` consumer uses only `work_unit_id` and `disposition`.

### No regressions ✅
Only the watcher and its tests changed since a51f0cc. At head: `test_sprint_pr_watcher.py` 21 passed + 7 subtests; the PR's other touched test files (`test_sprint_cli`, `test_sprint_recovery`, `test_sprint_review_loop`, `test_sprint_snapshot_rebuild`, `test_verify_clean_clone`) 64 passed + 2 subtests. Prior six verified-clean items stand.

### New finding
**L3 (Low) — reactivation failure at re-subscribe is silent and unretried.** If the fresh observation triggered by a repeat `sc pr subscribe` on a terminal-latest subscription fails (`GitHubReadError`, e.g. rate limit), no transition is recorded, so the subscription stays quiescent — and because quiescent rows are excluded from pulses, the backoff never gets a retry (a *fresh* subscribe self-heals: no transitions yet → still polled). The failure is durably recorded in `pr_subscription_poll_failures`, but the CLI receipt (`subscription_id`, `created`) doesn't surface it, so the operator believes observation resumed. Recovery: re-run the same idempotent command. Bounded resilience gap on a rare path — non-blocking; fix boundary if taken up: surface observation failure in the subscribe receipt, or let the registration observation's backoff entry re-qualify the row for pulse retry.

Recommendation: merge-ready from review's side (FnB's gate). Not merged, not pushed.

Issue #978, opened 2026-08-03
supervision tests: restricted restart intermittently omits pre-restart backup


jedbjorn:
Where: subfloor commit 2be3606 plus isolated Sprint U7 arming changes, DEV4 host seat, fresh linked worktree.

Ran: ./sc test

Expected: RestrictedLaunchTests.test_restart_bounces_systemd_broker_and_reports_full_inventory creates exactly one shell_db.prerestart.*.db fixture backup.

Actual: backup_files was empty (0 != 1). All Sprint/U7 coverage passed; full result was 1 failed, 1713 passed, 1 skipped in 94.20s.

Workaround: rerun the single test to determine whether this is an isolated fixture race; no supervision code is touched by U7.

jedbjorn:
Confirmed stable on immediate isolated rerun: 1 failed in 0.93s with the same 0 != 1 backup assertion.

Issue #977, opened 2026-08-03
sprint tests: PR watcher emits missing pr_subscriptions table on clean base


jedbjorn:
Where: subfloor commit 2be3606, DEV4 host seat, fresh linked worktree after ./sc deps.

Ran: .venv/bin/pytest -q tests/test_sprint_work_dispatch.py tests/test_sprint_v2_domain.py tests/test_sprint_cli.py

Expected: focused baseline passes and CLI emits only its JSON response.

Actual: tests/test_sprint_cli.py::SprintCliApiTest::test_declare_plan_lifecycle_and_registration_surfaces fails because register-pr writes 'sprint-pr-watcher: pulse failed (no such table: pr_subscriptions)' before the JSON response; json.loads then raises JSONDecodeError. Summary: 1 failed, 43 passed, 2 subtests passed.

Workaround: none applied; U7 verification will isolate its arming-flow tests and report this pre-existing baseline separately.

jedbjorn:
Follow-up: tests/test_sprint_cli.py later passed alone (11 passed, 2 subtests), and the full 1,714-test run also passed that file. The original failure is therefore order-/isolation-sensitive rather than a deterministic clean-base failure.

PR #979, opened 2026-08-03
feat(sprints): gate dispatch on arming selections

jedbjorn:
## Outcome

- validate every participant launch selection before lifecycle publication or work dispatch; null model/effort remain the explicit harness/default selections
- enqueue one declared-New wake to the originating Planner with the recorded Planner/Dev/Reviewer routes, then release eligible work in the same transaction
- expose Planner and work wake identities separately in the lifecycle event while retaining the complete initial wake list
- update delivery/liveness/live-proof coverage for the Planner chat opened by arming and later Re-enter traffic

## UI boundary

The assigned base has no browser arming action/modal; participant route pickers feed the existing declaration records and the Sprint board only exposes pause/resume/abort. This change reuses those stored harness/model/effort fields and adds no UI surface.

## Verification

- affected Sprint matrix: 96 passed
- Sprint CLI: 11 passed, 2 subtests passed
- Planner Re-enter live proof: passed
- targeted ruff, py_compile, node syntax, and diff checks: passed
- full suite: 1713 passed, 1 skipped, 1 unrelated repeatable supervision-backup failure (issue #978)
- order-sensitive PR-subscription test contamination captured in issue #977; the CLI file and full-suite CLI segment pass in isolation

## Trade-off

The model gate treats null model or effort as a recorded default, matching existing launch semantics; blank explicit strings fail loudly before any wake or disposition changes.

### DIFF
--- a/.super-coder/scripts/sprint_domain.py
+++ b/.super-coder/scripts/sprint_domain.py
@@ -255,23 +255,31 @@ class SprintLifecycleStore:
                 )
             self._require_edge(sprint["lifecycle"], "armed")
             self._authorize(sprint, "armed", actor)
-            planner_participant = self._validate_arm_plan(sprint)
+            planner_participant, selections = self._validate_arm_plan(sprint)
             self._update_lifecycle(
                 sprint_id,
                 current="prepared",
                 target="armed",
                 outcome=None,
             )
-            wake_ids = self._release_initial_work(
+            planner_wake_id = self._queue_arming_planner_wake(
+                sprint_id,
+                planner_participant_id=planner_participant,
+                selections=selections,
+            )
+            work_wake_ids = self._release_initial_work(
                 sprint_id,
                 planner_participant_id=planner_participant,
             )
+            wake_ids = [*work_wake_ids, planner_wake_id]
             self._event(
                 sprint_id,
                 "lifecycle.armed",
                 actor,
                 {
                     "initial_wake_ids": wake_ids,
+                    "planner_wake_id": planner_wake_id,
+                    "work_wake_ids": work_wake_ids,
                 },
             )
         return wake_ids
@@ -1805,7 +1813,10 @@ class SprintLifecycleStore:
                     "only an assigned participant may pause this Sprint"
                 )
 
-    def _validate_arm_plan(self, sprint: sqlite3.Row) -> int:
+    def _validate_arm_plan(
+        self,
+        sprint: sqlite3.Row,
+    ) -> tuple[int, list[sqlite3.Row]]:
         sprint_id = int(sprint["sprint_id"])
         if int(sprint["merge_grant_enabled"]) != 1:
             raise SprintInvariantError("arming requires a committed merge grant")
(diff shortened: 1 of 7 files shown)


jedbjorn (commented):
## S87 U7 adversarial review — arming flow + model-selection gate (spec #89 §Arming Flow, task #266) @ 0da35ba

**Counts: 0 Major / 0 Medium / 3 Low. Non-gating.**

### Deviation verdict: **deviated-intentionally — faithful given base reality; residual gap Low**

Verified against base `2be3606`: `ui/app.js` contains **no arming modal and no per-role pickers** — sprint actions are only Pause/Resume/Abort, and arming is CLI-only. The spec's "re-use the existing per-role pickers from the arming modal — no new UI surface" rests on a false premise at base: there is nothing to re-use, and providing a picker surface would have required exactly the new UI the spec forbade.

The implemented reading preserves every executable clause of §Arming Flow:
- dispatch gated on selections *recorded* — `_validate_arm_plan` rejects blank (non-NULL) model/effort for **any** participant before the lifecycle flip;
- NULL model/effort = defaults and **satisfy** the gate (main dispatch tests arm with NULL fields and pass, rendering "default" in the planner body);
- selections live in the existing per-participant `harness/model/effort` fields — the PR only *reads* them; no parallel store;
- arm always queues a declared-New wake to the overseeing Planner carrying every role's harness · model · effort.

What is lost is the FnB adjust-in-chat-before-dispatch affordance ("chat opens with pickers… adjust or say go"): work releases in the same transaction as the gate, so there is no adjustment window at arm time. Severity **Low**: selections are FnB-controllable at declaration, defaults are the spec's own zero-friction case, and the planner receives the full selection list at arm. Recommend a spec-#89 amendment (planner-owned) recording that arm-time selection is declaration-backed until a real picker surface exists — not a dev fix.

### Verified (traced, not trusted)
1. **Gate cannot be bypassed.** `arm()` is the only prepared→armed path: `transition()` raises "prepared Sprints must use arm()", `resume()` raises the same, `dispatch_ready()` returns `[]` unless lifecycle is already `armed`. Validation runs before `_update_lifecycle`; validation failure rolls back lifecycle, wakes, messages, dispositions together (tests assert `prepared`, `COUNT(wake_message)=0`, units still `planned`).
2. **Atomic release + real evidence separation.** Planner wake, work release, and `lifecycle.armed` event share one `sprint.arm` write transaction. `planner_wake_id` vs `work_wake_ids` is queryable beyond the event payload: the rows differ in `message_kind` (`notification` vs `work_assignment`) and the planner message carries a deterministic idempotency key.
3. **Declared-New planner wake.** `declared_type="new"` flows through the U2 delivery-time resolver (idle → close-first fresh chat; mid-turn → Re-enter under the delivery rule; coalesced any-new-rotates CASE intact). Tests confirm the arming delivery opens the planner registry chat and later planner-bound relays re-enter it.
4. **Idempotency is safe.** `sprint:{id}:arming-model-selections` cannot collide: arm is once-per-sprint (abort terminal, resume refuses prepared), and `_send`'s replay path raises on any input drift. The `receipt.wake_id is None` guard is unreachable belt-and-braces — `_send` either creates/coalesces a wake or raises.
5. **v2.0 interview superseded, no dual path.** No interview code path exists at base (`git grep -i interview` hits only the unrelated fork-naming text in `init_fork.py`/`install.py`); nothing competes with the gate. Skill/boot-doc wording is task #268 scope by build order.

### Findings (all Low)
- **L1 (conformance/deviation)** — the verdict above: adjust-before-dispatch affordance unimplemented; spec premise false at base. Route: spec amendment via planner; optionally a future UI unit adds the pickers and an arm-time adjustment window.
- **L2 (code quality, `sprint_domain.py::_validate_arm_plan`)** — the invariant error names no offender: "arming requires recorded model selections for every participant" while `invalid_selections` holds the exact rows. Enumerating role/shortname/field would make the fail-loud actionable without a DB dive. (Adjacent nit: the blank-`harness` clause is dead code — schema `CHECK (trim(harness)<>'')` + NOT NULL already forbid it, and `str(None)` would slip a hypothetical NULL through as `"None"`; the DB constraint is what actually holds the line.)
- **L3 (test coverage)** — no test exercises the arming wake against a planner with a **pre-existing idle-open chat** (spec: "closes planner's own chat if idle-open — intended") or a **mid-turn planner** (falls to the delivery rule). Both ride U2 generic machinery and are tested there, but the arming-specific claim in the spec is asserted nowhere.

Pre-existing scope excluded per assignment (#959/#960/#962/#966/#967/#972/#973/#977/#978, flag #160). No migration in this PR — none needed; correct. Checks green at head.

**Recommendation: mergeable as-is** (PLN1 gate, decision #77) with L1 routed to the planner as a spec amendment; L2/L3 are take-or-leave polish, fit for a later sweep.

PR #981, opened 2026-08-03
feat(sprints): retire legacy chat machinery

jedbjorn:
## Summary

- retire participant chat pointers, purpose/parent topology, reroute helpers, terminal lifecycle chat closure, and the Sprint-armed PR-subscription gate
- make active_shell_chats the sole current-chat authority, retain flat immutable Sprint history links, and enforce one open chat per shell universally
- fold Sprints v2.1 authority and delivery guidance into five skills plus boot, with convergent reseed migration 0169
- make full-suite skill and supervision fixtures hermetic

Fresh writer inventory at base d1b0aa7: 9 pointer-writing SQL sites. Remaining after this change: 0.

Trade-off: the API still exposes current_conversation_id as a derived registry projection for UI compatibility, while all mutable participant pointer/topology state is deleted.

## Verification

- ./sc test — 1717 passed
- migration/reseed slice — 80 passed, 199 subtests passed
- ./sc render-check
- ./sc verify
- git diff --check

Sprint 87 U9 · task #268. Never merge without the FnB gate.

### DIFF
--- a/.super-coder/api/server.py
+++ b/.super-coder/api/server.py
@@ -73,7 +73,7 @@ import sprint_liveness  # noqa: E402  (Sprints v2 one-shot monitor surface)
 import sprint_message_delivery  # noqa: E402  (Sprints v2 inbox acceptance)
 import sprint_recovery  # noqa: E402  (Sprints v2 pause/resume reconciliation)
 import sprint_review_loop  # noqa: E402  (Sprints v2 Dev/Review command surface)
-import sprint_runtime  # noqa: E402  (armed-only Sprint dispatch + wake delivery)
+import sprint_runtime  # noqa: E402  (Sprint dispatch + engine wake delivery)
 import sprint_board  # noqa: E402  (read-only Sprints v2 FnB board projections)
 import skill_projection  # noqa: E402  (exact bounded grant mirrors)
 sys.path.insert(0, str(ENGINE / "api"))
@@ -83,8 +83,8 @@ import map_db  # noqa: E402  (read-only handle to the dr_* catalogue in map.db)
 import ports as ports_mod  # noqa: E402
 import shell_factory  # noqa: E402
 import snapshot as snapshot_mod  # noqa: E402  (engine_skill_names — origin rule)
-import sprint_participant_chats  # noqa: E402  (Sprints v2 participant chat topology)
-import sprint_pr_watcher  # noqa: E402  (Sprints v2 armed GitHub observation)
+import sprint_participant_chats  # noqa: E402  (registry-backed Sprint wake chats)
+import sprint_pr_watcher  # noqa: E402  (engine-wide PR subscription observation)
 import model_catalog  # noqa: E402  (live model-id suggestions, sibling module)
 import analytics  # noqa: E402  (token & session analytics sweep — doc #11)
 import token_parsers  # noqa: E402  (harness roster + per-parser data dirs)
@@ -4294,7 +4294,7 @@ def require_loopback_bind(bind: str) -> None:
 
 
 def start_runtime_services() -> None:
-    """Start commit-woken conversations and the armed Sprint services."""
+    """Start commit-woken conversations and Sprint/engine services."""
     broker = conversation_broker.start_service(
         DB_PATH,
         launch_preparer=conversation_launch.ConversationLaunchPreparer(DB_PATH),
--- a/.super-coder/assets/skills/sprint_close/SKILL.md
+++ b/.super-coder/assets/skills/sprint_close/SKILL.md
@@ -22,6 +22,14 @@ question, answer, blocker, or context message, run `accept` for that message.
 For informational messages it only marks the message read; it does not change
 Sprint or work-unit state.
 
+Planner-bound close/conformance messages are Re-enter wakes resolved through
+the active-chat registry. A verified live turn is never displaced; delivery
+waits for its natural boundary and drains every undelivered message. An idle
+Re-enter resumes the registry chat, while coordinate mode makes it an idle New
+ticket chat after FnB closes the Planner chat. Automatic pause preserves that
+mode; FnB pause/resume returns to supervise. The inactivity ceiling and reaper
+own silent or unlinked processes, not this skill.
+
 ## Questions, answers, blockers, and failures
 
 Put a concrete question, answer, blocker, or context request in a short body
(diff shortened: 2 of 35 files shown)


jedbjorn (commented):
## S87 U9 review — REV3 (adversarial, spec doc #89 · task #268) @ a6b9bde

**Verdict: 0 Major / 0 Medium / 1 Low — merge gate satisfied.** Comment-only review; merge is PLN1's under decision #77.

### 1. Retirement residue — all 9 verified residue-free (independent grep at head)
- **Pointer columns**: independent base inventory confirms the dev's count — **9 pointer-writing SQL sites at d1b0aa7** (8 in `sprint_participant_chats.py`, 1 in `snapshot.py`), **0 at head**. Remaining mentions are the declared read-only derived projections (`active.chat_id AS current_conversation_id` in board/participant projections + API/UI), immutable migration history (0146/0147), and legacy-floor fixtures inside migration tests. No writer survives.
- **Purpose taxonomy / parent topology**: `_PURPOSES`, purpose-classified links, `parent_conversation_id`, context packets — gone from code; 0168 rebuilds the link table flat; `sprint_close` history projection updated.
- **Reroute chains + legacy raw-key dual lookups**: `linked_conversation_for_key`, `_reroute_operation_key`, `ensure_wake_conversation` deleted; no residue (the surviving "reroutes" test name covers new delivery-time re-resolution, not legacy chains).
- **`close_for_terminal_lifecycle`**: deleted, and both live call sites (transition-to-completed, abort) removed with it — no dangling callers anywhere at head. Abort still interrupts active runs via the immutable history-link join (`_persist_interrupt_intents`), pointer-free.
- **Message `active` flag / divergent producers / hand-rolled watcher cancellation**: all clean; remaining `sprint_wake_outbox` INSERTs are enqueue-only sites feeding the one coalescing outbox + single delivery-time resolution service.
- **Armed-only gate**: registration-side armed check removed in this PR (watcher polling gate fell in U6); new test asserts paused-sprint registration works.
- **Skill/doc/text residue**: zero hits for retired concepts across `assets/skills/`, `docs/`, `templates/`, `hooks/`, `adapters/`, UI.

### 2. Kept-list — intact, untouched
`idx_sprint_wake_one_pending_participant` (0148→0164, no type column), wake↔message join, `sprint_wake_attempts` evidence, acceptance triggers (0164), `conversation_generation` keying, hash-chain transition dedupe (watcher), queued-turn delivery — all present; their owning modules (`sprint_message_delivery.py`, 0148/0164) are not in this diff. 0168's rebuilt link table keeps `conversation_id UNIQUE` + immutability triggers + insert validation.

### 3. Test-migration honesty — clean
Deleted tests cover deleted machinery only (purposes, reroute keys, pointer rollback, fix/merge lanes, planner fallback packets, arming provisioning). Replacements assert the replacement behavior: registry registration + flat history, idempotent replay without duplicate history, committed-close-before-create, link immutability + cross-shell rejection, zero-chat-legal projection, completion-does-NOT-close-or-interrupt, abort-still-interrupts. API fixtures honestly rebuild sprint chats via registry inserts respecting one-open. **Independently executed** `test_sprint_participant_chats` + `test_sprint_recovery` + `test_sprint_snapshot_rebuild` + `test_sprint_message_delivery` (72 passed) and `test_sprint_skills` + `test_conversation_schema` (40 passed) at a6b9bde in a detached worktree.

### 4. Fold-in fidelity — complete; U8 Lows L1–L4 all fixed
Registry/delivery-rule/reaper/ceiling/coordinate-dial/subscriptions/arming-gate folded into all five role skills + `templates/boot.md` (new ACTIVE CHAT DELIVERY block with the exact three-row delivery table from spec #89). Verified against msg #1345: **L1** fixed (compile-report block no longer interleaves mid-sentence), **L2** fixed (`replan` added to the Reviewer never-runs list), **L3** fixed (sprint_dev:154 now "Reviewer decides… Planner executes"), **L4** fixed (sprint_close pause doctrine now Rev-decides/Planner-executes + decision #46 override). Catalogue-wide sweep: no stale v2.0 doctrine remaining in any skill. Arming text matches the U7 amendment (validation gate, defaults satisfy, no phantom pickers).

### 5. Three-artifact integrity — proven by execution
0169 parity test derives expected values by parsing the asset SKILL.md files, applies 0169 **twice** onto a deliberately dirtied 0168-floor DB — idempotent replay + dirty-row convergence + asset↔migration byte parity all proven (ran it myself, green). `./sc render-check` passes in the worktree. Numbering unique: origin/main tip is 0167; this PR adds 0168/0169. `sprint_removal` manifest updated to carry both + boot.md.

### 6. Schema drops — upgrade-safe on live-value DBs
0168 drops the 0147 pointer triggers and indexes **before** the column drops (SQLite would otherwise refuse); `DROP COLUMN` precedent exists (0097). The dev's upgrade test executes 0168 against a DB with live pointer values, a dirty purposed link, and an active registry row — columns dropped, history preserved row-exact, `PRAGMA foreign_key_check` clean (executed, green). The universal `idx_conversations_one_open_shell` is safe against the multi-open states the old exempted index permitted because 0162 (same upgrade batch for downstream installs) scope-blindly closes all but the newest open chat per shell after cancelling their queued work; for installs running interim code, every runtime creation path (registry close-before-create) and the reopen route (auto-close/refuse, `conversation_routes.py:474-510`) enforce one-open, so no two-open row can survive to 0168. Nothing FK-references the renamed link table, so the RENAME/DROP rebuild cannot retarget foreign keys.

### Findings
**L1 (Low, report-only)** — `sprint_domain.py:1633` `participant_capacity.available` semantics drifted: it now means "shell not deleted AND has a registry chat". Since chats are created only at first wake delivery (arming no longer provisions), a legally idle participant (zero registry chat — v2.1 ground state; e.g. right after arming, after an FnB close, or a coordinate-mode planner between tickets) reports `available: 0` in the resume-reconciliation projection. No consumer gates on it (report-only), but the label can mislead a Planner reading the reconciliation report into treating idle as broken. Cheap fix whenever convenient: rename the field (e.g. `has_active_chat`) or drop the chat term from the predicate.

Note (not a finding): sprint pill/board/liveness projections surface the shell's registry chat even when it is not sprint-scoped (e.g. an FnB chat that displaced the sprint chat). Read as intended under spec #89's universality doctrine — the registry is the current-chat authority.

Ruled scope honored: #959/#960/#962/#966/#967/#972/#973/#977/#978, flag #160, map gap #1347 not re-raised.

— REV3 (review shell, different lineage than author). Never merging, never pushing.

PR #990, opened 2026-08-03
feat(sprints): delivery-terminal reviewer wake + compile authority

jedbjorn:
## What shipped

- Task 274: added the transactional `sprint.delivery_terminal` event and durable Reviewer notification when the last work unit becomes completed/cancelled through merge observation, report/no-code completion, or cancellation.
- Task 275: added resume-time catch-up after re-arming, including merge observations projected while paused.
- Task 276: allowed participating Reviewers to compile the read-only evidence packet while preserving Planner/FnB access and every other close authority.
- Task 277: added the delivery-terminal and compile-authority matrix beside the existing Sprint lifecycle suites.

The existing board event allowlist now projects the three delivery-terminal count fields. No endpoint, page, schema, table, daemon, or skill asset was added.

## Verification

`/home/j3d1/Repos/sc-cachy/.venv/bin/python -m pytest tests -q`

Result: **1729 passed, 1 skipped, 876 subtests passed in 111.90s**.

## Edge cases covered

- final merge after pause → resume → serial work units
- final report-only/no-code completion
- cancellation completing an all-cancelled set
- no wake mid-Sprint, while paused, or for a grant-bypassed merge
- resume-time all-terminal catch-up
- same-count replay dedupe and fresh-key re-fire after added scope
- one notification per participating Reviewer, with exact body/type/actionability and system event payload
- Reviewer, Planner, and FnB compile access; Developer and non-participating Planner refusal

## Judgment calls

- `wake_message.idempotency_key` is globally unique. A single-Reviewer Sprint uses the canonical `sprint:<id>:delivery-terminal:<terminal_count>` key exactly; multi-Reviewer delivery retains that namespace and appends `:reviewer:<participant_id>` so every Reviewer receives a distinct durable message without a schema change.
- The full-suite event-emitter contract required adding `sprint.delivery_terminal` to the existing board projection allowlist. This exposes only the three count fields and does not create or change an endpoint.

### DIFF
--- a/.super-coder/scripts/sprint_board.py
+++ b/.super-coder/scripts/sprint_board.py
@@ -33,6 +33,9 @@ _EVENT_FIELDS = {
     "sprint.declared": frozenset(
         {"feature_id", "spec_approval_ids", "participant_shell_ids"}
     ),
+    "sprint.delivery_terminal": frozenset(
+        {"terminal_count", "completed_count", "cancelled_count"}
+    ),
     "lifecycle.armed": frozenset(
         {
             "from",
--- a/.super-coder/scripts/sprint_close.py
+++ b/.super-coder/scripts/sprint_close.py
@@ -251,7 +251,9 @@ class SprintCloseStore:
             raise ValueError(
                 f"section limit must be between 1 and {MAX_SECTION_LIMIT}"
             )
-        sprint = self._require_close_authority(sprint_id, caller_shell_id)
+        sprint = self._require_close_authority(
+            sprint_id, caller_shell_id, allow_reviewer=True
+        )
         events = self._events(sprint_id)
         specs = self._spec_revisions(sprint_id)
         units = self._planned_vs_actual(sprint_id)
@@ -541,7 +543,11 @@ class SprintCloseStore:
         return events
 
     def _require_close_authority(
-        self, sprint_id: int, caller_shell_id: int
+        self,
+        sprint_id: int,
+        caller_shell_id: int,
+        *,
+        allow_reviewer: bool = False,
     ) -> sqlite3.Row:
         row = self.con.execute(
             "SELECT sp.*,r.title AS feature_title,s.flavor AS caller_flavor "
@@ -555,6 +561,10 @@ class SprintCloseStore:
             int(row["originating_planner_shell_id"]) != caller_shell_id
             and row["caller_flavor"] != "admin"
         ):
+            if not allow_reviewer:
+                raise SprintAuthorityError(
+                    "only the owning Planner or FnB may record the final report"
+                )
             try:
                 self._require_reviewer(sprint_id, caller_shell_id)
             except SprintAuthorityError:
--- a/.super-coder/scripts/sprint_close.py
+++ b/.super-coder/scripts/sprint_close.py
@@ -555,7 +555,13 @@ class SprintCloseStore:
             int(row["originating_planner_shell_id"]) != caller_shell_id
             and row["caller_flavor"] != "admin"
         ):
-            raise SprintAuthorityError("only the owning Planner or FnB may compile")
+            try:
+                self._require_reviewer(sprint_id, caller_shell_id)
+            except SprintAuthorityError:
+                raise SprintAuthorityError(
+                    "only the owning Planner, FnB, or a participating Reviewer "
+                    "may compile"
+                ) from None
         return row
 
     def _require_participant_or_admin(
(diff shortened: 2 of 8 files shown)


jedbjorn (commented):
**REVIEW VERDICT: REQUEST CHANGES** (posted as comment — the gh token owns this PR, so GitHub refuses a formal request-changes review) — S88 U1 adversarial review @ d555a671, against spec-94 + decisions #82/#45/#67. Full notes: `shared/sprints/sprint-88/u1-pr990-review.md` (flag SC-990-FINAL-REPORT).

**1. Major — compile-authority widening leaks into `record_final_report`** (`.super-coder/scripts/sprint_close.py:129`, via the widened `_require_close_authority` at :543-565). The guard has two callers; the diff widened the guard itself, so a participating Reviewer now passes authority for the **final synthesis**, not just compile. Spec-94 Engine Changes §3 is explicit: `record_final_report` is *untouched*; this breaks the #67 lanes (Rev decides, Planner acts). Failure scenario: a Reviewer records a final report mid-conformance; when the Planner executes the conclude decision, `record_final_report` with the Planner's body raises `SprintInvariantError` ("Sprint already has a different final report") — there is no amend/delete path for a final report row, so the conclude path is wedged short of DB surgery. Compounding: a reviewer-authored final report is stamped `actor_kind="planner"` in the `final_report.recorded` event (:156), misattributing authorship. No test pins reviewer-refused on `record_final_report`, so the suite stays green with this leak. Fix direction: scope the reviewer branch to compile only (e.g. `allow_reviewer` param defaulted False, passed True from `compile_evidence_packet`) + a negative test.

**2. Low — `arm()` never checks the all-terminal state** (`.super-coder/scripts/sprint_domain.py:242`). Units can be cancelled while `prepared`; arming an all-cancelled Sprint releases no work and no call path ever reaches `_queue_delivery_terminal` — the Sprint sits armed-and-done silently, the exact stall this spec exists to kill. Resume got a catch-up check; arm did not. Pathological, but the "all units cancelled: the wake still fires" edge-case promise only holds when the last cancellation happens while armed.

**3. Low — stale test name** (`tests/test_sprint_close.py:718`): `test_only_planner_or_admin_compiles_and_participants_read_timeline` now asserts the Reviewer compiles. Rename to match behavior.

Verified clean: trigger fires exactly once per episode across all three disposition paths (merge-observed, report_only/no_code `complete`, `cancel`); armed-only guard re-queried in-helper; grant-bypass cannot false-trigger; resume catch-up correctly placed post-re-arm and skipped on re-pause; count-keyed idempotency + per-reviewer suffix matches spec; wake shape (`notification`/`new`/actionable=0) honors #65/#66/#81 and rides the existing outbox/recovery ladder; board allowlist addition is projection-only with no consumer break; tests bite on lifecycle guard, dedupe, resume call, key derivation, and bypass negatives (the `changed` guard at sprint_domain.py:2368 is unpinned but harmless — the helper self-guards).

Requesting changes on finding 1; findings 2 and 3 are advisory.

jedbjorn:
## REV1 re-review (r2) — head `1401ccf0`

**Verdict: APPROVE** — 0 Critical / 0 Major / 0 Medium / 0 Low. All three r1 findings verifiably fixed; full notes appended to `shared/sprints/sprint-88/u1-pr990-review.md` (r2 section).

**(Major, r1 #1 — fixed, pin bites)** `_require_close_authority` now defaults `allow_reviewer=False`; only `compile_evidence_packet` opts in (sprint_close.py:254). `record_final_report` (sprint_close.py:129) refuses a participating Reviewer before any write — the r1 wedge scenario (reviewer-authored final → planner conclude dies on `SprintInvariantError`) is closed, and the `actor_kind="planner"` misattribution is unreachable. New test `test_participating_reviewer_cannot_record_final_report` uses shell 2, a participating reviewer of *this* sprint, and asserts zero report rows + zero events — re-widening turns it red. Bites.

**(Low, r1 #2 — fixed, no side effects)** `arm()` catch-up (sprint_domain.py:285) runs after the in-transaction lifecycle update to `armed`, inside the same write tx, using the same store idiom as the resume path. Helper guards no-op normal arms (pending units → not-all-terminal) and dedupe by `terminal_count`; pre-arm fire is impossible, so no double-fire. `test_arming_all_cancelled_sprint_wakes_reviewers` pins recipient shells [2, 5], one event, canonical episode key. Bites.

**(Low, r1 #3 — fixed)** Test renamed to `test_planner_admin_and_participating_reviewer_compile_evidence`; acceptance fixture entry updated, and the manifest harness asserts referenced test names exist — machine-checked, not cosmetic.

Regression sweep: production delta is exactly these two hunks; all r1-verified surfaces (terminal helper body, complete/cancel/merge/resume call sites, board allowlist, wake shape) untouched; CI green at reviewed head.

Clean to merge.

Issue #995, opened 2026-08-03
tests: PR watcher pulses after fixture database teardown


jedbjorn:
Where: subfloor source checkout, DEV3 host seat.

Ran: `.venv/bin/pytest -q tests/test_sprint_board_api.py tests/test_sprint_message_delivery.py tests/test_sprint_work_dispatch.py`

Expected: green test completion with fixture-owned watcher services stopped before database teardown.

Actual: pytest exits 0 with 72 passed / 14 subtests, then reproducibly prints `sprint-pr-watcher: pulse failed (no such table: pr_subscriptions)`.

Impact: green suite emits a real-looking runtime failure and reveals a daemon/test teardown race; it can obscure verification evidence.

Workaround: none needed for test outcome; treat the late line as teardown noise. Spec #96 does not touch watcher lifecycle.

PR #993, opened 2026-08-03
feat(skills): closeout loop protocol + sprint artifact paths

jedbjorn:
## Summary
- teach Reviewers to enter closeout from the delivery-terminal wake and branch between in-Sprint re-entry and record/conclude
- teach Planners to materialize Reviewer re-entry decisions while keeping compile-report as an FnB-directed fallback
- standardize gitignored Sprint artifact paths across Reviewer, Planner, Close, and Developer skills
- reseed the four role skills and add focused content/idempotency coverage

## Verification
- `./sc test` — 1736 passed, 1 skipped
- `./sc render-check` — passed

## Judgment
- refreshed `0001_seed_skills.sql` per the source-repo migration-management contract and added trailing migration 0171 for existing installs
- delivery-terminal notifications with non-terminal units are treated as stale; all-cancelled delivery resolves to abort

## Tooling note
- filed #991 after observing that `./sc seed-skills --help` executes the seeder instead of showing help

PR #996, opened 2026-08-03
feat(sprints): honor sanctioned quiet liveness

jedbjorn:
## Summary

- replace recursive Sprint wake text with precise durable-write retry guidance
- classify pending owned PRs and outbound handoffs as liveness suppressors, with failure evidence retaining precedence
- emit once-per-episode sanctioned-quiet evidence and a transition-keyed 90-minute Planner CI backstop
- reseed the Developer skill with once-only pre-handoff inbox guidance and a named native-wake pause state

## Verification

- `./sc test` — 1748 passed, 1 skipped
- `./sc render-check` — pass
- focused Ruff — pass
- isolated mypy on changed source modules — pass

## Notes

- 5-minute evaluation cadence, 10-minute grace, and 10-minute escalation window are unchanged.
- Out-of-scope reproducible test teardown noise filed as #995.

### DIFF
--- a/.super-coder/migrations/0001_seed_skills.sql
+++ b/.super-coder/migrations/0001_seed_skills.sql
@@ -3939,6 +3939,12 @@ failure paths. A local exploratory number is not merge evidence. Record real CI
 failures, anomalous infrastructure failures, retries, review friction, and
 known departures for the final report.
 
+Immediately before a typed Developer handoff (`complete-unit`, `register-pr`,
+or `request-review`), re-run `sc sprint inbox --sprint <id>` once and act on
+anything new; a ruling may have arrived during the build. This is a once-only
+pre-handoff check. After the handoff confirms its durable write, stop without a
+further inbox pass.
+
 An explicitly planned report-only or no-code lane completes with its durable
 result instead of a PR. Code lanes cannot use this path; they complete only
 after merge authorization and observation.
@@ -3975,8 +3981,8 @@ registered PR.
 Complete a review handoff in this exact order:
 
 1. Finish the readiness claim and every local verification step.
-2. Re-run `sc sprint inbox --sprint <id>`, act on newly arrived messages, and
-   mark every handled informational message read with `accept`.
+2. Perform the once-only typed-handoff inbox check above, act on newly arrived
+   messages, and mark every handled informational message read with `accept`.
 3. Run `wc -m < <path>`; keep the claim near 6,000 characters and below the
    8,000-character hard maximum.
 4. As the literal final action of the turn, send the typed handoff with one
@@ -4044,7 +4050,8 @@ the Reviewer decides whether to continue, re-plan, or pause and the Planner
 executes that decision.
 
 Stop when the unit is merged and reported, declined, awaiting Planner/FnB
-recovery, or returned to review. For normal review and merge handoffs, the
+recovery, returned to review, or paused awaiting a native PR-fact or verdict
+wake. For normal review and merge handoffs, the
 ordered procedures above place inbox handling before the typed handoff and make
 that handoff the turn''s last action. Ask the Planner for later work only after
 the current editing lane is terminal.',
--- a/.super-coder/migrations/0172_sanctioned_pause_liveness.sql
+++ b/.super-coder/migrations/0172_sanctioned_pause_liveness.sql
@@ -1,11 +1,16 @@
----
-name: sprint_dev
-description: Execute a Sprints v2 Developer lane — accept one assignment, implement and verify it, own the PR through green and review, merge only after live authorization, and record judgment without overlapping edits.
-category: workflow
-common: false
----
+-- 0172 — reseed sanctioned-pause Developer guidance.
+-- Full-body UPSERT converges existing installations to spec #96's once-only
+-- pre-handoff inbox check and named native-wake pause state.
 
-# sprint_dev — own one editing lane
+BEGIN;
+
+INSERT INTO skills (name, description, category, command, common, content, is_deleted) VALUES (
+  'sprint_dev',
+  'Execute a Sprints v2 Developer lane — accept one assignment, implement and verify it, own the PR through green and review, merge only after live authorization, and record judgment without overlapping edits.',
+  'workflow',
+  NULL,
+  0,
+  '# sprint_dev — own one editing lane
 
 Use for an actionable work-unit assignment in an armed Sprint. Marking that
 Sprint message read is acceptance and starts work immediately. If you cannot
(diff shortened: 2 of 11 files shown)


jedbjorn:
## Review — spec #96 (Sanctioned pause): REQUEST-CHANGES

1 Major / 0 Medium / 2 Low. CI green at e9b21aa; review is against the spec's contracts, traced through the code at this head.

### Major — the `outbound.handoff` suppressor cannot engage in production

`.super-coder/scripts/sprint_liveness.py:395-400` admits the outbound suppressor only when the participant's latest `wake_message.created_at` is **strictly after** `last_strong_at`. But `collect()` (`sprint_liveness.py:184-187`) computes `last_strong_at = max(stored, strong.observed_at)`, and `_latest_strong` counts every native chat event in `_NATIVE_EVIDENCE_EVENTS` (`sprint_liveness.py:28-39` — including `tool.completed` and `run.completed`) as strong evidence.

The handoff send (`sc sprint send` / `request-review` / verdict) executes as a tool call inside the participant's registered sprint chat. The `wake_message` row is written *during* that command; the harness then records the call's `tool.completed`, the closing `assistant.delta`, and the turn's `run.completed` — all seconds **after** the row's `created_at`. The next evaluation stores the send turn's completion as `last_strong_at`, so `created_at > last_strong_at` is permanently false. Result: the suppressor never holds, and Rev/Planner — plus any Developer handoff without a pending PR (e.g. a question to the Planner) — are nudged ~10 minutes after stopping, exactly the loop spec #96 and decision #85(3) ("rev/planner = activity OR outbound durable send since last strong evidence") set out to kill. The spec's verification item "Suppressor: outbound send … → no nudge" is unachievable in production; only the PR-pending suppressor (no timestamp comparison) actually works.

The tests miss it because they inject sends via `SprintMessageStore.send` and backdate `created_at` with an `UPDATE` (`tests/test_sprint_liveness.py:699-718`) without emitting the send's own `tool.completed` — and the collector test at `tests/test_sprint_liveness.py:743-763` itself demonstrates that one `tool.completed` after the send releases the suppressor. Production always has one.

Suggested direction (yours to design): evaluate suppression against strong evidence *outside the send's own turn* — e.g. treat the send turn's completion events as part of the handoff, or compare the message against strong evidence predating it — and add a test that reproduces the real event ordering (message row, then `tool.completed`/`run.completed` in the same chat).

### Low — suppressors are participant-scoped, muting unrelated expectations

`_suppressors` (`sprint_liveness.py:363-411`) collects per-participant state and `_apply` (`sprint_liveness.py:693`) applies it to whichever expectation is being evaluated. A pending PR therefore also silences a second live expectation of the same participant (e.g. an accepted-but-unanswered Planner question) whose next-wake obligation was never transferred — in tension with the spec's stated principle ("sanctioned when the participant's last act was a durable handoff"), though the role-instantiation text ("nudge only if … no PR under CI") sanctions it. Flagging for a ruling: intended, or expectation-scoped?

### Low — CI backstop only fires while the PR owner has a live expectation

`_send_ci_stalled_backstops` is reachable only from `_apply`'s suppressor branch (`sprint_liveness.py:693-694`), i.e. during evaluation of an unresolved expectation. If all of the owner's expectations are resolved while the PR re-enters pending (post-`in_review` re-push awaiting a fresh review round), 90 minutes of stalled CI produces no Planner wake — the spec states the backstop unconditionally. Narrow, and partially covered by the reviewer's own expectation path, but worth naming.

### Verified clean against the spec's contracts

- **No retiming:** `EVALUATION_INTERVAL`/`GRACE_WINDOW`/`ESCALATION_WINDOW` byte-unchanged, plus an explicit regression test.
- **Precedence:** the failure branch now evaluates before suppressors (`sprint_liveness.py:647-658`); the reorder is semantics-preserving vs. the old fresh-strong ordering, and `process.missing` + pending PR still escalates (tested).
- **Durable-only evidence:** suppressors read `sprint_registered_prs` ownership + confirmed `wake_message` sends; `transition_key` is globally UNIQUE, so the backstop idempotency key cannot collide; transition insert is state/head-sha-gated, so the 90-minute clock is stable and resets on transition as specified.
- **Once-per-episode:** `liveness.sanctioned_quiet` is keyed on (expectation, episode); repeats update silently (tested).
- **Backstop:** one Planner wake per transition key, dev never contacted, mandate text matches spec (tested end-to-end).
- **Reach:** `sprint_board.py` additions are payload allow-list entries for the two new events within the existing event-stream projection — no new table, endpoint, or daemon; no UI change.
- **Two-artifact skill edit:** asset + reseed 0172 + seed 0001 converge hermetically (catalogue-parity test + render-check in CI); 0001 edit matches the PR #993 doctrine precedent; removal-manifest ledger is a frozen pre-removal baseline, correctly untouched.
- **Wake text:** loop sentence gone for every role, retry sentence present, covered for all three roles plus delivery/dispatch templates; no-suppressor silence still nudges and escalates exactly as today (existing fake-clock tests + new release-after-green test).

### Nits (no action requested)

- The once-only paragraph names `complete-unit`/`register-pr`/`request-review`, but the post-merge `send` handoff keeps its own separate re-run wording — two parallel phrasings of one rule, future drift risk.
- `liveness.sanctioned_quiet` can be emitted inside the grace window (first evaluation where a suppressor holds) — harmless, matches "first suppression of an episode".

jedbjorn:
**RE-REVIEW @ 8ee56497 — APPROVE** (Rev-03, scoped to round-1 findings)

**Major (flag #173) — fixed.** `_outbound_handoff` now anchors on the durable send's containing run end: latest outbound `wake_message` (`created_at >= accepted_at AND message_id > expectation_message_id`), containing `conversation_runs` turn resolved via `started_at <= sent_at <= ended_at`, release check `strong.observed_at > turn.ended_at`. Verified the contract both ways:
- (a) handoff + its own same-turn tool.completed/run.completed/run.heartbeat (≤ turn end) => suppression HOLDS, strong nulled — the round-1 prod masking is gone;
- (b) genuine later-run activity (strong strictly after turn end) => suppressor returns None, strong preserved, normal evaluation resumes.
- Regression test `test_outbound_handoff_ignores_its_turn_then_releases_on_new_run` injects exactly the realistic send-plus-same-turn-events sequence found masked in round 1 (run [5,7], handoff@6, tool.completed@6 + run.completed@7 bound to that run_id), then releases on a new run [11,12].

**Low 2 — fixed.** CI backstop moved to the top of `evaluate()`, sprint-scoped and independent of live expectations; idempotency key `liveness:ci-stalled:{transition_key}:planner` unchanged => still once-per-transition. New test proves it fires with zero live expectations; the existing once-per-transition test still pins dedup, green-release, and re-stall re-backstop. `sprint_board.py` field whitelist updated in sync with the payload change (only two consumers). New unconditional write-transaction is safe — both `evaluate()` callers invoke it outside transactions.

**Low 1 — landed.** `_suppressors` docstring cites decision #85(3); the PR suppressor query is byte-identical to round 1 — no behavior change.

**Round-1 clean list — no regressions.** `_apply` precedence untouched (failure > fresh strong > suppressor > nudge); quiet path still retimes only on the EVALUATION_INTERVAL cadence, `last_strong_at` untouched by suppressors; suppressor evidence remains durable-only.

CI on 8ee56497: 6/6 green (independently verified).

Non-blocking nits for the record: second-granularity stamps mean a later run's event landing in the exact same second as turn end holds suppression one extra cycle (self-corrects next tick); a send-turn run that never closes (`ended_at` NULL forever) never suppresses — fails noisy toward nudge/escalation, which is the safe direction.

Closing flag #173. Merge is the FnB's gate.

PR #1014, opened 2026-08-04
fix(sprint): honor live turn evidence in delivered-unread wake recovery

jedbjorn:
## Bug

Duplicate wake messages during Sprints: the same `wake_message` was delivered twice (or more) into the same chat. Observed live in the sc-cachy install's Sprint 1 — the Planner received `wake_message #50` twice with identical content, plus repeated empty "transport wakes" (see wakes 57→59, 60→61, 62→63 in that install's engine DB).

## Root cause

`_reconcile_unread_wakes_in_transaction` (sprint_domain.py) recovers wakes whose messages are delivered-but-unread. It computes `_wake_turn_evidence(old_wake_id)` — which correctly sees the just-delivered turn — but never consults it for Sprint-scoped wakes. Its only liveness guard is `_participant_has_pickup_turn`, which joins through `sprint_participant_conversations`.

The Originating Planner's chat predates the Sprint and is never registered in that table, so the guard is permanently blind for the Planner (and for any wake delivered before a participant conversation is registered). Sequence, repeated for every Planner wake:

1. Wake delivers into the Planner's active chat; turn enqueued (`queued`/`running`).
2. Next 5-second pulse: message is delivered-but-unread, pickup guard sees nothing → wake declared lost.
3. Reconciler resets `delivered_at=NULL`, mints `sprint-recovery:<sprint>:delivered-unread:<wake>` with a **fresh idempotency key**, re-points the messages.
4. The recovery wake re-delivers the identical prompt into the same chat — the enqueue dedupe can't catch it because the key differs.

The `sprint-recovery:` prefix skip stops a third round, so each wake doubles exactly once. The reconciler even recorded the evidence against itself: all twelve `wake.requeued` events in Sprint 1 carry `prior_turn_state` of `queued` or `running`.

## Fix

`_wake_turn_evidence` now renders a `turn_live` verdict — latest attempt's turn is `queued`, or `running` with a live un-interrupted run, in an open conversation — using the same liveness criteria `_participant_has_pickup_turn` already applies, but anchored on the wake's actual target conversation instead of the registration table. The delivered-unread branch skips recovery while `turn_live` holds.

Recovery semantics preserved:
- turn terminally finishes (succeeded/failed) with the message still unread → one recovery wake fires, as before;
- error-state conversation with a stranded queued turn → still recovers (`test_error_conversation_queued_turn_does_not_suppress_recovery` unchanged);
- SHELL_BUSY chain and failed-wake resume paths unchanged (a failed run is never `turn_live`).

## Tests

Two regressions in `tests/test_sprint_recovery.py` reproduce the Planner shape (active chat with no `sprint_participant_conversations` row): queued-turn and running-turn variants, both red on main, green here. Full suite: 1815 passed, 907 subtests.

### DIFF
--- a/.super-coder/scripts/sprint_domain.py
+++ b/.super-coder/scripts/sprint_domain.py
@@ -1160,6 +1160,11 @@ class SprintLifecycleStore:
                 and turn["run_state"] == "succeeded"
             ):
                 continue
+            # The delivered turn is still queued or actively running: the
+            # receiver has not had its chance to read yet.  Requeuing now
+            # duplicates the wake into the same chat under a fresh key.
+            if turn["turn_live"]:
+                continue
             shell_busy = (
                 turn["run_state"] == "failed"
                 and turn["error_code"] == "SHELL_BUSY"
@@ -1463,7 +1468,9 @@ class SprintLifecycleStore:
             (participant_id,),
         ).fetchone() is not None
 
-    def _wake_turn_evidence(self, wake_id: int) -> dict[str, str | int | None]:
+    def _wake_turn_evidence(
+        self, wake_id: int
+    ) -> dict[str, str | int | bool | None]:
         wake = self.con.execute(
             "SELECT idempotency_key FROM sprint_wake_outbox WHERE wake_id=?",
             (wake_id,),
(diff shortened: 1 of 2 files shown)


jedbjorn:
**Cross-fork verification — dos-arch (true fork, independent engine DB).**

Its `sprint_wake_outbox` holds 25 `delivered-unread` recovery wakes across Sprints 3 and 5. Joining each prior wake's delivery attempt against `sprint_participant_conversations` and the recorded `wake.requeued` turn evidence partitions them exactly:

| target conversation | turn evidence at requeue | count | verdict |
|---|---|---|---|
| unregistered | `queued` / `running` / `leased` (live) | 21 | illegitimate duplicate |
| registered | `completed` / `cancelled` / `failed` (terminal) | 4 | legitimate recovery |

No mixed cases. The 21 duplicates: PLN1's chat `cv_0652…` (Sprint 5, 15×), PLN2's `cv_bd30…` (Sprint 3, 5×), and REV2's `cv_1ce1…` (Sprint 3, 1×) — the reviewer case shows the exposure isn't planner-only: any shell whose active chat isn't a Sprint-registered conversation (pre-Sprint chat, or post-rotation) is hit. The `turn_live` guard in this PR keys on the actual target conversation, so it suppresses exactly the 21 and preserves all 4 legitimate recoveries.

PR #1420, opened 2026-08-30
U1: add private instance-state resolver

jedbjorn:
Work unit: 141
Spec: #186

### DIFF
--- a/.super-coder/api/conversation_routes.py
+++ b/.super-coder/api/conversation_routes.py
@@ -25,7 +25,6 @@ from urllib.parse import parse_qs, urlparse
 
 ENGINE = Path(__file__).resolve().parents[1]
 REPO_ROOT = ENGINE.parent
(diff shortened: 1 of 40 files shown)


jedbjorn:
Rev-01 review — merge blocker at exact head `ca6a5b1922b9`:

All `instance.json` writers must share the resolver’s locked, atomic update path. The new canonical updater is used by the installer, but `ports.py`, `feature.py`, and the `db-init` / `pg-init` paths still rewrite the same file directly without `_instance_id_lock` or atomic replacement. A concurrent or interrupted write can corrupt or remove the supposedly immutable `instance_id`; `ports.py` additionally treats malformed JSON as empty and can overwrite it, allowing a later resolver call to allocate a different private namespace.

Please route every `instance.json` mutation through the canonical updater and add regressions for concurrent writers and malformed configuration, then re-push for review.

PR #1639, opened 2026-09-21
fix(models): refresh never records host harness versions for a sandbox install

jedbjorn:
## Problem

Flag #653 item 4 (dos-app e2e `p2a_report.md`, Medium). On a sandbox-runtime install, an unauthenticated host `./sc models refresh` fell through to the direct-DB `model_catalog.catalog(refresh=True)` path and recorded the HOST's harness CLI versions as route evidence. Launch binds evidence to the runtime that executes the route, so the API rejected with `thinking_evidence_stale` ("Installed harness version changed after refresh") and named `sc models refresh` as the remedy — a loop the remedy could not escape.

## Fix

Invariant: versions recorded as evidence come from the runtime the install launches shells in.

- `dispatch.sh` — new `sc_models`: an unauthenticated host-seat `models refresh` on a sandbox install execs into the running container (`docker exec $CNAME python3 $S/models.py refresh`), the same seam `harness-status` already probes through. In-container it runs the identical `catalog(refresh=True, con=con)` the GUI's `GET /api/models?refresh=1` runs. Sandbox down / no docker: exit 2, names `./sc launch, then ./sc models refresh`, records nothing.
- `models.py` — `_require_launch_seat()`: a direct-DB refresh on a sandbox install from a non-sandbox seat (anything that bypasses the dispatcher) refuses before opening the DB or probing.
- Unchanged: host-runtime installs, in-sandbox seats, token seats (already reach the engine API in that runtime), and `list` / `resolve`.

No doc or skill text became wrong — `./sc models refresh` remains the documented command.

## Tests

`python3 -m pytest tests/test_model_catalog.py tests/test_harness_epoch.py tests/test_host_runtime.py tests/test_host_runtime_contract.py tests/test_shell_alias.py tests/test_callable_floor.py tests/test_route_bindings.py tests/test_instance_state.py tests/test_worktree_targets.py -q` — 285 passed.

New: host seat + sandbox install records nothing (models.py); sandbox seat records; host install unchanged; dispatcher execs into the sandbox, refuses when it is down, stays on host for a host-runtime install, and leaves `list` and token refresh on the calling seat.

Not verified here: a live container run (no live engine touched by design). Inference rests on the in-container path being the same code the API endpoint runs.


jedbjorn:
Review round 1 addressed in 6a2c7289 + 72085f55: `resolve` now routes through the same sandbox exec seam (argv forwarded, exit status propagated) and is covered by the direct-module seat guard; unreachable-runtime refusals are split by cause (no docker / daemon down / container stopped) and exit 3, distinct from models.py's 2; fake docker exec can exit non-zero and the propagated status is asserted; README headless sections and harness-freshness.md carry the sandbox clause. `list` stays on the calling seat — it only reads rows, no cross-seat comparison.

